PR #117, opened 2015-01-05
Api vcs

pypingou:


puiterwijk:
Awesome, I hadn't known that str(dict) is an almost-json object!

Looks good to me, :+1:

pypingou:
Thanks for the review!

PR #118, opened 2015-01-05
Start optimize /api/bugzilla by generating the JSON output ourself

pypingou:


pypingou:
Before this change: ``12538441 function calls (9658724 primitive calls) in 10.732 seconds``
With this change: ``6286887 function calls (6286711 primitive calls) in 9.013 seconds``

pkgdb2/api/extras.py
```diff
@@ -216,7 +218,10 @@ def api_bugzilla():
     acls = _bz_acls_cached(name, out_format)
 
     if out_format == 'json':
-        return flask.jsonify(acls)
+        return flask.Response(
+            str(acls).replace("'", '"').replace('None', 'null'),
```
ralphbean: Usage of ``str`` here is ambiguous, no?  Perhaps instead you want ``unicode(acls).encode('utf-8')``?

ralphbean: Also, this seems error prone in general.

- Did you try ``import json; json.dumps(acls)``?  Was it faster than ``flask.jsonify(acls)``?
- What if there is a single ``'`` in some string?  Replacing it with ``"`` would be incorrect.
- When converting the string to json, true you need to convert ``None`` to ``null``, but you also need to convert ``True`` to ``true`` and ``False`` to ``false``.

pypingou: On Wed, Jan 07, 2015 at 08:23:42AM -0800, Ralph Bean wrote:
>    In pkgdb2/api/extras.py:
> 
>  > @@ -216,7 +218,10 @@ def api_bugzilla():
>  >      acls = _bz_acls_cached(name, out_format)
>  > 
>  >      if out_format == 'json':
>  > -        return flask.jsonify(acls)
>  > +        return flask.Response(
>  > +            str(acls).replace("'", '"').replace('None', 'null'),
> 
>    Usage of str here is ambiguous, no? Perhaps instead you want
>    unicode(acls).encode('utf-8')?

I will try if this works

pypingou: On Wed, Jan 07, 2015 at 08:25:56AM -0800, Ralph Bean wrote:
>    In pkgdb2/api/extras.py:
> 
>  > @@ -216,7 +218,10 @@ def api_bugzilla():
>  >      acls = _bz_acls_cached(name, out_format)
>  > 
>  >      if out_format == 'json':
>  > -        return flask.jsonify(acls)
>  > +        return flask.Response(
>  > +            str(acls).replace("'", '"').replace('None', 'null'),
> 
>    Also, this seems error prone in general.
> 
>      * Did you try import json; json.dumps(acls)? Was it faster than
>        flask.jsonify(acls)?

That's what flask.jsonify is doing:
https://github.com/mitsuhiko/flask/blob/master/flask/json.py#L201

>      * What if there is a single ' in some string? Replacing it with " would
>        be incorrect.

Since the output only contains package name, FAS username and FAS groups, I
think we're safe on this.

>      * When converting the string to json, true you need to convert None to
>        null, but you also need to convert True to true and False to false.

jsonlint.com seems to agree with you, I wonder why the unit-tests passed then
:-/

pypingou:
> jsonlint.com seems to agree with you, I wonder why the unit-tests passed then 

And the answer is simple, there is no True/False (Boolean) in the output :-)

ralphbean:
:+1: :)

pypingou:
Thanks for the review!

PR #120, opened 2015-01-12
Retired pops-up information as why the retirement (according to dead.package)

pypingou:


pkgdb2/api/extras.py
```diff
@@ -585,3 +586,29 @@ def api_monitored():
...
+    Retired packages should have in their git a ``dead.package`` file
+    containing the explanation as why the package was retired.
+    This method calls cgit to return that explanation.
+
+    ::
+
+        /api/dead/package/acheck/master
+
+    '''
+    req = requests.get(
```
ralphbean: A thought:  would it make more sense to return a ``302`` redirect here instead getting the content ourselves?

pypingou: Ok I tried to do this, but since this API is designed to be called via jquery/ajax, the 302 triggers the cross-site security which also prevents us from calling cgit directly.

If you have an idea to call cgit directly (or redirect to it and having jquery happy about it), feel free :-)

ralphbean: Nope.  :+1: as is.  Thanks!

PR #122, opened 2015-01-12
Go back to rely on flask.jsonify to generate the JSON output in the API

pypingou:
Trying to generate the JSON manually ends up with some invalid JSON as
some packages' summary or description contain some " or some ' which
messes up the JSON we generate.


puiterwijk:
Too bad it didn't work, code looks good to me.

pypingou:
Thanks for the review

PR #123, opened 2015-01-12
Generate the JSON for the /api/vcs endpoint at the right place

pypingou:


ralphbean:
Easy :+1:.  Makes sense.

pypingou:
Thanks !

PR #124, opened 2015-01-13
Add missing join condition

puiterwijk:

### DIFF
--- a/pkgdb2/lib/model/__init__.py
+++ b/pkgdb2/lib/model/__init__.py
@@ -1788,6 +1788,8 @@ def get_groups(session):
     ).filter(
         PackageListing.collection_id == Collection.id
     ).filter(
+        PackageListingAcl.packagelisting_id == PackageListing.id
+    ).filter(
         PackageListing.status == 'Approved'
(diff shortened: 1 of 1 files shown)


ralphbean:
@puiterwijk can you explain some what's going on here?

ralphbean:
:+1: after discussion in IRC.   If it it hot-fixed and working now, perhaps @pypingou should have a look too before we merge to develop.

puiterwijk:
The problem is that on the join of PackageListing with PackageListingAcl, there is no join condition.

To give numbers:
PackageListing currently has 218804 records
PackageListingAcl currently has 1735450 records.
This means that without a join condition (which is what I'm adding here), it tries to find distinct rows between 3011786702500 rows.
Which is... quite a few :).

puiterwijk:
This change reduced running time of the query from about 2 minutes down to about 10 seconds.

pypingou:
Thanks for the fix @puiterwijk I was suspecting something like this but had not looked at it yet.

Merging so that I can make a .2 release later today

PR #130, opened 2015-01-19
Turns out `@patch('pkgdb2.lib.utils.get_packagers')` does not work

pypingou:
We can patch the whole module and set a return_value for one method in it
but patching the method directly doesn't work.

So we need to create some logs manually to test the timeline endpoint

### DIFF
--- a/tests/test_flask_ui_packages.py
+++ b/tests/test_flask_ui_packages.py
@@ -496,12 +496,13 @@ class FlaskUiPackagesTest(Modeltests):
                 '<li class="errors">No package of this name found.</li>'
                 in output.data)
 
-    @patch('pkgdb2.lib.utils.get_packagers')
+    @patch('pkgdb2.lib.utils')
     @patch('pkgdb2.packager_login_required')
     def test_package_take(self, login_func, mock_func):
         """ Test the package_take function. """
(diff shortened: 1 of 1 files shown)


tests/test_flask_ui_packages.py
```diff
@@ -496,12 +496,13 @@ def test_package_retire(self, login_func, mock_func):
...
                 in output.data)
 
-    @patch('pkgdb2.lib.utils.get_packagers')
+    @patch('pkgdb2.lib.utils')
     @patch('pkgdb2.packager_login_required')
     def test_package_take(self, login_func, mock_func):
         """ Test the package_take function. """
         login_func.return_value = None
-        mock_func.return_value = ['pingou', 'toshio']
+        mock_func.get_packagers.return_value = ['pingou', 'toshio']
```
ralphbean: FWIW, ``mock_func`` should probably be renamed to something like ``utils_module`` or ``utils_mod``.

pypingou: Yes it should, in the whole file :-)

ralphbean:
:+1:

pypingou:
Thanks @ralphbean for the review

PR #131, opened 2015-01-21
Admin actions

pypingou:
This is a pretty big changeset.
We basically add a workflow for packager to request new packages or new branches for existing packages via pkgdb2 (instead of relying on bugzilla.redhat.com as we do now).

The idea is:
* After having had the review on bugzilla, the packager asks for the package to be added in pkgdb
* Admins review the request and grant or deny it (in case of denial, a reason must be provided)
* The package is created in pkgdb, git is automatically sync using https://github.com/fedora-infra/fedmsg-pkgdb-sync-git

* Packager asks for a new branch on an existing package
* Branch for Fedora releases are automatically granted
* Branch for EPEL go through the following process
 * For 1 week the request is in `Pending` state
 * While the request is in `Pending` state, the current packager with `approveacls` on the package can block the request (they will have to provide a reason for this) or accept it (ie: change its status to `Awaiting Review`).
 * Requests `Awaiting Review` are processed by admins who will either grant it or deny it (again they will have to provide a reason to do so)
* Upon request for a new branch in EPEL, pkgdb2 will check if the package is present in RHEL and if so if it is present on all arches or just some, in the former case the request will be blocked, in the later the request will be recorded.

In addition of these changes in the UI, these changeset provide the corresponding API to be used by a new tool ``pkgdb-admin`` developed as part of `packagedb-cli`.


createdb.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/python
 
 ## These two lines are needed to run on EL6
-__requires__ = ['SQLAlchemy >= 0.7', 'jinja2 >= 2.4']
+__requires__ = ['SQLAlchemy >= 0.8', 'jinja2 >= 2.4']
```
ralphbean: do we need to worry about el6 anymore?  Can these lines be removed?

pkgdb2/api/admin.py
```diff
@@ -0,0 +1,355 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © 2014  Red Hat, Inc.
```
ralphbean: Update copyright year while you're here?

pkgdb2/forms.py
```diff
@@ -127,11 +139,11 @@ class AddPackageForm(wtf.Form):
         'Summary',
         [wtforms.validators.Required()]
     )
-    description = wtforms.TextField(
+    description = wtforms.TextAreaField(
         'Description',
     )
     review_url = wtforms.TextField(
-        'Review URL',
+        'Review URL <span class="error">*</span>',
```
ralphbean: Is this ``<span>`` not added automatically by ``wtforms`` when you add the ``wtforms.validators.Required()``?

pkgdb2/lib/__init__.py
```diff
@@ -291,7 +335,8 @@ def set_acl_package(session, pkg_name, pkg_branch, pkg_user, acl, status,
         raise PkgdbException('No collection found by the name of %s'
                              % pkg_branch)
 
-    if not pkgdb2.is_pkg_admin(session, user, package.name, pkg_branch):
+    if not force and not pkgdb2.is_pkg_admin(
```
ralphbean: ``force`` seems like a bad idea.

So, anyone can create ACLs without being an admin?

pkgdb2/lib/__init__.py
```diff
@@ -1529,6 +1644,238 @@ def add_branch(session, clt_from, clt_to, user):
     return messages
 
 
+def add_new_branch_request(session, pkg_name, clt_from, clt_to, user):
+    """ Register a new branch request.
+
+    This method only flushes the new objects.
```
ralphbean: This part of the docstring doesn't make sense.  Could you elaborate more?

pkgdb2/lib/model/__init__.py
```diff
@@ -1590,6 +1641,191 @@ def insert(cls, session, user, package, description):
...
+    )
+
+    @property
+    def info_data(self):
+        """ Return the dict stored as string in the database as an actual
+        dict object.
+
+        """
+        if self.info:
+            return ast.literal_eval(self.info)
```
ralphbean: Whoah.  ``literal_eval``?  I see that the docstring claims it is 'safe', but I've only ever seen ``json`` used for situations like these.

pkgdb2/lib/model/__init__.py
```diff
@@ -1090,6 +1121,26 @@ def by_name(cls, session, pkgname):
         """
         return session.query(cls).filter(Package.name == pkgname).one()
 
+    @property
+    def requests_pending(self):
+        """ Returns the list of pending branch requests
+        """
+        requests = []
```
ralphbean: could this be made tighter with a list comprehension?

createdb.py
pypingou: On Wed, Jan 21, 2015 at 07:45:21AM -0800, Ralph Bean wrote:
>    In createdb.py:
> 
>  > @@ -1,7 +1,7 @@
>  >  #!/usr/bin/python
>  > 
>  >  ## These two lines are needed to run on EL6
>  > -__requires__ = ['SQLAlchemy >= 0.7', 'jinja2 >= 2.4']
>  > +__requires__ = ['SQLAlchemy >= 0.8', 'jinja2 >= 2.4']
> 
>    do we need to worry about el6 anymore? Can these lines be removed?

We do not have to since we are running EL7 but I do not know if rpmfusion is
running 6 or 7 and they expressed interest in pkgdb. So for the moment I prefer
to remain EL6 compatible :)

pkgdb2/lib/__init__.py
pypingou: On Wed, Jan 21, 2015 at 07:52:20AM -0800, Ralph Bean wrote:
>    In pkgdb2/lib/__init__.py:
> 
>  > @@ -291,7 +335,8 @@ def set_acl_package(session, pkg_name, pkg_branch, pkg_user, acl, status,
>  >          raise PkgdbException('No collection found by the name of %s'
>  >                               % pkg_branch)
>  > 
>  > -    if not pkgdb2.is_pkg_admin(session, user, package.name, pkg_branch):
>  > +    if not force and not pkgdb2.is_pkg_admin(
> 
>    force seems like a bad idea.
> 
>    So, anyone can create ACLs without being an admin?

I agree that forces is a bad idea, but since new branch request for Fedora
branches are automatically approved and all the person needs to make such
request is being in the packager group, I needed a way to automatically grant
that person its ACLs without intervention from an admin.

So the only place where force is used is:
https://github.com/fedora-infra/pkgdb2/blob/admin_actions/pkgdb2/lib/__init__.py#L1739

pkgdb2/lib/__init__.py
pypingou: On Wed, Jan 21, 2015 at 07:54:15AM -0800, Ralph Bean wrote:
>    In pkgdb2/lib/__init__.py:
> 
>  > @@ -1529,6 +1644,238 @@ def add_branch(session, clt_from, clt_to, user):
>  >      return messages
>  > 
>  > 
>  > +def add_new_branch_request(session, pkg_name, clt_from, clt_to, user):
>  > +    """ Register a new branch request.
>  > +
>  > +    This method only flushes the new objects.
> 
>    This part of the docstring doesn't make sense. Could you elaborate more?

Actually it can be removed as it does not even ``session.flush()``, all it does
it ``session.add``. The idea was that there is nothing committed to the database
after this method has been called.

pkgdb2/lib/model/__init__.py
pypingou: On Wed, Jan 21, 2015 at 07:57:17AM -0800, Ralph Bean wrote:
>    In pkgdb2/lib/model/__init__.py:
> 
>  > +        "Collection",
>  > +        foreign_keys=[collection_id], remote_side=[Collection.id],
>  > +    )
>  > +    from_collection = relation(
>  > +        "Collection",
>  > +        foreign_keys=[from_collection_id], remote_side=[Collection.id],
>  > +    )
>  > +
>  > +    @property
>  > +    def info_data(self):
>  > +        """ Return the dict stored as string in the database as an actual
>  > +        dict object.
>  > +
>  > +        """
>  > +        if self.info:
>  > +            return ast.literal_eval(self.info)
> 
>    Whoah. literal_eval? I see that the docstring claims it is 'safe', but
>    I've only ever seen json used for situations like these.

Well, self.info is actually a JSON-like object. I wonder if I could not have
used json.load() directly

pkgdb2/lib/__init__.py
ralphbean: OK.  I understand.

pkgdb2/lib/model/__init__.py
ralphbean: Yeah.  Using eval and the ``ast`` on data in the database seems like a vulnerability waiting to happen (although I can't think of a way to specifically exploit it as written).  I'd be more comfortable using ``json.{loads,dumps}`` if we can.

pkgdb2/api/admin.py
```diff
@@ -0,0 +1,355 @@
...
+
+from math import ceil
+
+import pkgdb2.lib as pkgdblib
+import pkgdb2.forms
+from pkgdb2 import SESSION, is_admin
+from pkgdb2.api import API, get_limit
+
+
+@API.route('/admin/actions/')
```
puiterwijk: Perhaps /admin/actions/pending or /admin/pending is more descriptive?

pkgdb2/api/admin.py
```diff
@@ -0,0 +1,355 @@
...
+    ::
+
+        /api/admin/actions/
+
+    Accept GET queries only.
+
+    :kwarg package: restrict the actions to a specific package.
+    :kwarg packager: restrict the actions to a specific packager.
+    :kwarg action: restrict the actions to a specific action, options are:
+        ``request.branch``.
```
puiterwijk: Maybe list all valid options, as you do for status?

pkgdb2/api/admin.py
```diff
@@ -0,0 +1,355 @@
...
+              "action": "request.branch",
+              "collection": {
+                "branchname": "epel7",
+                "dist_tag": ".el7",
+                "koji_name": "epel7",
+                "name": "Fedora EPEL",
+                "status": "Active",
+                "version": "7"
+              },
+              "date_created": 1402470695.0,
```
puiterwijk: Why do you return float's for timestamps?
I would say integers have enough precision as I wouldn't need to know microseconds of when the request was put in.

pkgdb2/api/admin.py
```diff
@@ -0,0 +1,355 @@
...
+
+import pkgdb2.lib as pkgdblib
+import pkgdb2.forms
+from pkgdb2 import SESSION, is_admin
+from pkgdb2.api import API, get_limit
+
+
+@API.route('/admin/actions/')
+@API.route('/admin/actions')
+def api_admin_actions():
```
puiterwijk: Shouldn't this have is_admin?

pkgdb2/api/admin.py
```diff
@@ -0,0 +1,355 @@
...
+        output['page_total'] = 1
+
+    jsonout = flask.jsonify(output)
+    jsonout.status_code = httpcode
+    return jsonout
+
+
+@API.route('/admin/action/<actionid>')
+@API.route('/admin/action/')
+def api_admin_action(actionid=None):
```
puiterwijk: is_admin here as well?

pkgdb2/api/admin.py
pypingou: Well, it does not return only pending actions

pkgdb2/api/admin.py
pypingou: That was meant to be this way as you can see, but there is an option missing indeed

pkgdb2/api/admin.py
pypingou: That's not something I handle and I'd rather leave it the way python creates it :)

pkgdb2/api/admin.py
pypingou: Well, this is read-only, and I don't think there is anything confidential there, after all the requests are sent via fedmsg so they are already public

pkgdb2/api/admin.py
pypingou: This is also read-only and not really confidential

pkgdb2/api/admin.py
```diff
@@ -0,0 +1,355 @@
...
+        status=action_status,
+    )
+    if form.validate_on_submit():
+        action_id = form.id.data
+
+        admin_action = pkgdblib.get_admin_action(SESSION, action_id)
+        if not admin_action:
+            output['output'] = 'notok'
+            output['error'] = 'No Admin action with this identifier found'
+            httpcode = 500
```
puiterwijk: Isn't this just a 404, or at least a 4xx code (request error)?
It's definitely not a server-error, as 5xx indicates.

pkgdb2/api/admin.py
```diff
@@ -0,0 +1,355 @@
...
+    else:
+        output['output'] = 'notok'
+        output['error'] = 'Invalid input submitted'
+        if form.errors:
+            detail = []
+            for error in form.errors:
+                detail.append('%s: %s' % (error,
+                              '; '.join(form.errors[error])))
+            output['error_detail'] = detail
+        httpcode = 500
```
puiterwijk: This sounds like 400 ("Bad request") to me.

pkgdb2/api/admin.py
pypingou: Problem is the API uses 500 for pretty much everything that is not 200 and I have no way of checking is a client has hard-coded the check for 500 or not.

pkgdb2/api/admin.py
pypingou: Same as for the 500, just trying to be consistent with the current mis-use of http error code

pkgdb2/api/admin.py
```diff
@@ -0,0 +1,355 @@
...
+    ..note:: The ``Obsolete`` status can only be set by the person who made
+             the request.
+
+    Sample response:
+
+    ::
+
+        {
+          "output": "ok",
+          "messages": ["Admin action status updated to: Approved"]
```
puiterwijk: Again with the full sentences in response codes?
I would prefer machine-interpretable responses so you could for example translate the API client without having to do manual string fiddling.

pkgdb2/lib/__init__.py
```diff
@@ -1529,6 +1645,236 @@ def add_branch(session, clt_from, clt_to, user):
...
+    :arg pkg_description: the description of the package.
+    :arg pkg_status: the status of the package.
+    :arg pkg_collection: the collection in which had the package.
+    :arg pkg_poc: the point of contact for this package in this collection
+    :arg user: the user performing the action
+    :kwarg pkg_review_url: the url of the review-request on the bugzilla
+    :kwarg pkg_upstream_url: the url of the upstream project.
+    :kwarg pkg_critpath: a boolean specifying if the package is marked as
+        being in critpath.
+    :returns: a message informating that the request has been successfully
```
puiterwijk: informing is probably the word you want to use

pkgdb2/lib/__init__.py
```diff
@@ -1529,6 +1645,236 @@ def add_branch(session, clt_from, clt_to, user):
...
+    :arg session: session with which to connect to the database.
+    :arg pkg_name: the name of the package.
+    :arg pkg_summary: a summary description of the package.
+    :arg pkg_description: the description of the package.
+    :arg pkg_status: the status of the package.
+    :arg pkg_collection: the collection in which had the package.
+    :arg pkg_poc: the point of contact for this package in this collection
+    :arg user: the user performing the action
+    :kwarg pkg_review_url: the url of the review-request on the bugzilla
+    :kwarg pkg_upstream_url: the url of the upstream project.
```
puiterwijk: Would we really need this, as I think we update this from the spec file later anyway?

pkgdb2/api/admin.py
pypingou: >    Again with the full sentences in response codes?
>    I would prefer machine-interpretable responses so you could for example
>    translate the API client without having to do manual string fiddling.

We have had this discussion before and while I agree with you, for the moment I
keep the behavior consistent with the rest of the API

pkgdb2/lib/__init__.py
pypingou: >    Would we really need this, as I think we update this from the spec file
>    later anyway?

Well, it's kind of nice to know what a new package is about without waiting for
the Friday update to know more about it, don't you think ? ;-)

pkgdb2/lib/model/__init__.py
```diff
@@ -211,6 +223,25 @@ def all_txt(cls, session):
             session.query(cls).order_by(cls.status).all()]
 
 
+class ActionStatus(BASE):
+    ''' Table storing the statuses for the AdminActions. '''
```
puiterwijk: Why do you use ' instead of " here?

pkgdb2/lib/model/__init__.py
```diff
@@ -1590,6 +1643,191 @@ def insert(cls, session, user, package, description):
         session.flush()
 
 
+class AdminAction(BASE):
+    """This associates a package with a particular collection.
```
puiterwijk: I think this might be a copy-paste error, and should've said something else?

pkgdb2/templates/list_actions.html
```diff
@@ -0,0 +1,126 @@
...
+{% block content %}
+
+<h1>Actions</h1>
+
+{% if total_page and total_page > 1 %}
+<table>
+    <tr>
+        <td>
+        {% if page > 1%}
+            <a href="{{ url_for('.admin_action') }}?page={{page - 1}}&package={{package}}&from_date={{from_date}}">
```
puiterwijk: Why don't you use url_for('.admin_action', page=page-1, package=package, ...?
As per the documentation (http://flask.pocoo.org/docs/0.10/api/#flask.url_for), all unknown arguments are added as query parameters.

pkgdb2/lib/__init__.py
puiterwijk: Then why don't we trigger an update of this specific package as soon as it's requested?

pkgdb2/templates/list_actions.html
pypingou: We could, not sure what we gain from it and I find this syntax clearer in a way (you know what is and what is not supposed to be in the URL).
Just personal taste, I guess

pkgdb2/lib/__init__.py
pypingou: Because the package is not in git even less in yum's metadata ? :)

puiterwijk:
I would like the policy to be defined before this getting merged (so whether or not current package admins should approve new branch requests for Fedora as well), but the code looks okay to me.

puiterwijk:
After @pypingou explained he wanted to test it, I'm okay with getting merged and to staging, but I would like to wait with upgrading prod until a decision has been made.

pypingou:
Thanks for the reviews @puiterwijk and @ralphbean 
I will merge and make a pre-release (1.22.99) waiting for the final green light before prod.

PR #133, opened 2015-01-28
Requests page

pypingou:


ralphbean:
I don't understand the ``{{ select }}`` var in the templates.  What is it supposed to represent?

pypingou:
@ralphbean the ``{{ select }}`` precise on the template which item of the menu to highlight and I used this to mark the difference between listing requests for an user (select = 'packagers') and admin (select undefined).

ralphbean:
As mentioned in IRC, I'd also add a comment explaining the differences in the queries.

ralphbean:
Thanks.  :+1: here.

pypingou:
Thanks for the review @ralphbean

PR #134, opened 2015-01-31
Simpler form

pypingou:


pkgdb2/ui/packages.py
```diff
@@ -1130,14 +1130,24 @@ def package_request_branch(package, full=True):
...
 
-    collections = pkgdb2.lib.search_collection(
-        SESSION, '*', 'Under Development')
-    collections.extend(pkgdb2.lib.search_collection(SESSION, '*', 'Active'))
-    pkg_status = pkgdb2.lib.get_status(SESSION, 'pkg_status')['pkg_status']
+    collections = pkgdb2.lib.search_collection(SESSION, '*', 'Under Development')
+    collections.reverse()
+    active_collections = pkgdb2.lib.search_collection(SESSION, '*', 'Active')
+    active_collections.reverse()
+    # We want all the branch `Under Development` as well as all the `Active`
```
ralphbean: It would be much better to be able to retrieve the list of active branches from sort of releng web service... but that doesn't exist.  ;)  This will do.

tests/test_pkgdblib.py
```diff
@@ -1828,38 +1828,39 @@ def test_search_actions(self):
...
-            status=None,
-            page=1,
-            limit=50,
-        )
-        self.assertEqual(len(actions), 1)
-        self.assertEqual(actions[0].user, 'pingou')
-        self.assertEqual(actions[0].package.name, 'guake')
-        self.assertEqual(actions[0].collection.branchname, 'el6')
-        self.assertEqual(actions[0].from_collection.branchname, 'master')
+        if not os.environ.get('OFFLINE'):
```
ralphbean: What is this OFFLINE var?  Perhaps add a comment explaining it.

pkgdb2/ui/packages.py
pypingou: Actually this list is quite maintained as Under Development and active branches have different behaviors.

And technically, when Fn+1 is published, Fn-1 is still active for a month :)
So we would have to introduce a new status (pre-EOL?)

tests/test_pkgdblib.py
pypingou: It's mentioned in the commit message but indeed. Basically it allows skipping this test which requires network access (and thus allows having the test-suite pass even offline)

ralphbean: Cool.  Is it mentioned anywhere in the README/dev instructions?

I was thinking an inline comment might be helpful for future devs.. but a README note would be even better.

ralphbean:
:+1:

pypingou:
Thanks @ralphbean

PR #135, opened 2015-02-03
If the new user starts with 'group:' let's check it exists in FAS and is valid

pypingou:


pypingou:
This would prevent someone from giving watch* ACLs to an invalid group.

ralphbean:
Sure.  :+1:

PR #136, opened 2015-02-10
New configuration tweaking and documentation

pypingou:


ralphbean:
LGTM.  :+1:

doc/configuration.rst
```diff
@@ -226,3 +226,43 @@ package or something alike. These emails are sent to the address set
...
+
+
+Cookie conflicts
+----------------
+
+If you run multiple applications at different level of your server, by default
+the ``path`` of the cookie will be ``/``, eventually leading to cookie conflict
+but providing a working configuration out of the box
+
+To prevent this, adjust the ``APPLICATION_ROOT`` as needed.
```
puiterwijk: Maybe add that they should make sure the path starts with a forward slash.

utility/pkgdb2.cfg.sample
```diff
@@ -92,3 +92,15 @@ PKGDB2_EMAIL_CC = None
 ## package or something alike. These emails are sent to the address set
 ## here:
 #MAIL_ADMIN = 'admin@pkgdb'
+
+
+## List the packages that are not accessible to the provenpackager group
+PKGS_NOT_PROVENPACKAGER = ['firefox']
```
puiterwijk: This is not the same as what the documentation says as default: 'firefox' vs 'firefox', 'thunderbird', 'xulrunner'

pypingou: This is also not the default, this is a sample configuration file :)

doc/configuration.rst
pypingou: I'll add a note for this

I merged a bit too early :(, I had the page opened in a tab didn't see it was the one you commented on

PR #138, opened 2015-02-10
Make it easy to change your avatar.

ralphbean:
This makes it so that when you are logged in, anywhere your avatar
appears it will be clickable and will take you to libravatar so you can
change it.  Importantly, you'll be logged into libravatar correctly
which can sometimes be tricky.

### DIFF
--- a/pkgdb2/ui/__init__.py
+++ b/pkgdb2/ui/__init__.py
@@ -27,7 +27,7 @@ import flask
 
 import pkgdb2.lib as pkgdblib
 from pkgdb2 import (APP, SESSION, FAS, is_pkgdb_admin, __version__,
-                    is_safe_url)
+                    is_safe_url, is_authenticated)
 
 
 UI = flask.Blueprint('ui_ns', __name__, url_prefix='')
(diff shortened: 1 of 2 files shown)


pkgdb2/ui/__init__.py
```diff
@@ -58,14 +58,19 @@ def branches_filter(branches):
...
-    """ Template filter sorting the given branches, Fedora first then EPEL,
-    then whatever is left.
-    """
-    output = '<img class="avatar circle" src="%s"/>' % (
-        pkgdblib.utils.avatar_url(packager, size)
-    )
-
-    return output
+    """ Template filter to produce the libravatar of a given packager. """
+    if flask.g.fas_user and packager == flask.g.fas_user.username:
```
pypingou: There is an is_authenticated utility method that might be good to use here

ralphbean: Good call.

ralphbean: Fixed.

pkgdb2/ui/__init__.py
```diff
@@ -58,14 +58,19 @@ def branches_filter(branches):
...
-    then whatever is left.
-    """
-    output = '<img class="avatar circle" src="%s"/>' % (
-        pkgdblib.utils.avatar_url(packager, size)
-    )
-
-    return output
+    """ Template filter to produce the libravatar of a given packager. """
+    if is_authenticated() and packager == flask.g.fas_user.username:
+        openid_template = 'http://{packager}.id.fedoraproject.org'
```
puiterwijk: Note that the identity includes a trailing slash.

puiterwijk:
Looks good to me

pypingou:
I will add the trailing slash and merge as I would like to cut another release today.

PR #139, opened 2015-02-11
Add title to the avatar to explain what clicking on it does

pypingou:


puiterwijk:
That might be useful yeah. :+1:

PR #140, opened 2015-02-11
Add a placeholder for the Review URL of the package_request template

pypingou:


pkgdb2/templates/package_request.html
```diff
@@ -17,7 +17,8 @@
     {{ render_field_in_row(form.description) }}
     {{ render_field_in_row(form.branches) }}
     {{ render_field_in_row(form.review_url, title="URL pointing directly to the
- location where the package review happened") }}
+ location where the package review happened",
+     placeholder="https://bugzilla.redhat.com/1234") }}
```
ralphbean: This should probably be ``https://bugzilla.redhat.com/show_bug.cgi?id=1234``, no?

pypingou: @puiterwijk had a similar remarks, but the version here is also valid (and nicer imho).

ralphbean:
:+1:, then.  :)

PR #141, opened 2015-02-12
If the user does not have a bugzilla account, report it instead of failing

pypingou:


pkgdb2/lib/__init__.py
```diff
@@ -135,6 +135,23 @@ def _validate_pkg(session, rhel_ver, pkg_name):
                         pkg_name, rhel_ver))
 
 
+def _validate_fas_user(username):
+    """ Validate is the provided ``username`` is associated to a valid FAS
```
ralphbean: Should be "Validate that the provided ..."

ralphbean:
Aside from one cosmetic comment, :+1:.

pypingou:
Thanks!

PR #146, opened 2015-02-17
Drop branch from

pypingou:
When requesting a new branch we used to asked from which branch this new branch should be created.
However, the way we automated things, we cannot specify to the script from which branch to create the new one, instead we create the new branch pointing it to the first commit in the repo.

So there is no longer need for us to store this information in the DB nor to display it.


ralphbean:
Nice!  Simple is better than complex.  :+1:!

PR #147, opened 2015-02-17
Allow user to edit their own request

pypingou:


pkgdb2/ui/packages.py
```diff
@@ -349,9 +351,10 @@ def package_request_edit(package, action_id):
         )
 
     # Check user is the pkg/pkgdb admin
+    pkg_admin = pkgdblib.has_acls(
+        SESSION, flask.g.fas_user.username, package, 'approveacls')
```
ralphbean: How will this line work if ``package=None``?

pypingou: Local tests showed it behaves correctly but I've added a second check in has_acls itself to cover the situation (and avoid querying the DB for nothing)

ralphbean:
Cool.  :+1:.

pypingou:
Thanks @ralphbean

PR #158, opened 2015-02-20
Set monitor to True by default with new packages

puiterwijk:


puiterwijk:
#159 is the better one.

PR #159, opened 2015-02-20
Fix tests and setting the monitoring to true

pypingou:


pypingou:
Different approach to the same problem than in https://github.com/fedora-infra/pkgdb2/pull/158

puiterwijk:
Looks good to me :+1:

PR #162, opened 2015-02-23
Apply the correct patch to be able to check the return messages

pypingou:
Instead of patching the whole module, we only patch the specific method
we need to patch which thus allowing to check the messages returned by
the API


ralphbean:
Cool.  :+1:

PR #164, opened 2015-02-24
Fix excluding provenpackager from some packages specified in the configuration

pypingou:


ralphbean:
This change is being made to a very confusing piece of code.  From IRC:

```
threebean: what is 'pkg' ?
threebean: is it a list of fixed length?
   pingou: threebean: it is yes, retrieved from the db
threebean: what do the elements represent?
threebean: I ask because it's confusing to read with all the 'pkg[0]' and 'pkg[2]' stuff all over the place
threebean: like, "if pkg[2] not in output[pkg[0]]:"  who knows what that means..
threebean: can we change the vcs_acls code in pkgs2/lib/__init__ to read like
threebean: 'for package_name, username, branchname in pkgs:'
threebean: it's a lot more code change than is proposed in this PR
threebean: but I don't think I can actually review it :P  not without pen and paper
   pingou: I need to backport this in prod, so I'd like to keep this as small as possible
```

That said, the change looks sane enough, so :+1: as is.  I'll file another issue to clean up the code path.

Issue #165, opened 2015-02-24
Clean up pkgdb2.lib:vcs_acls


ralphbean:
It is *really* confusing.  See the comments in #164.

PR #166, opened 2015-02-25
Improve readability of pkgdb.lib.vcs_acls by using proper variable names

pypingou:
Fixes https://github.com/fedora-infra/pkgdb2/issues/165

### DIFF
--- a/pkgdb2/lib/__init__.py
+++ b/pkgdb2/lib/__init__.py
@@ -2025,6 +2025,134 @@ def bugzilla(session, name=None):
     return output
 
 
+def _vcs_acls_json(packages):
+    """ For a given list of package/user/branch build a dict of dict
+    representating of who has commit access to which package.
+
+    The output dict is something like:
+
+    {
+      pkg1: {
+        branch1: {
+          name: pkg1,
+          branch: branch1,
+          user: [user1, user2],
+          group: [group1, group2]
+        },
+        branch2: {
+          name: pkg1,
+          branch: branch2,
+          user: [user1],
+          group: [group1, group3]
+        },
+      },
+      pkg2:
+      ...
+    }
+    """
+    output = {}
+    for pkgname, username, branchname in packages:
+        user = None
+        group = None
+        if username and username.startswith('group::'):
+                group = username.replace('group::', '')
+        else:
+            user = username
+
+        if pkgname not in output:
+            output[pkgname] = {}
+
+        if branchname not in output[pkgname]:
+            groups = []
+            if skip_pp and pkgname not in skip_pp:
+                groups.append('provenpackager')
+
+            output[pkgname][branchname] = {
+                'commit': {'groups': groups, 'people': []}
+            }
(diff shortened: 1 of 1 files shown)


pkgdb2/lib/__init__.py
```diff
@@ -2038,75 +2038,75 @@ def vcs_acls(session, eol=False, oformat='text', skip_pp=None):
...
+            if username and username.startswith('group::'):
+                    group = username.replace('group::', '')
             else:
-                user = pkg[1]
+                user = username
 
-            if pkg[0] not in output:
-                output[pkg[0].encode('utf-8')] = {}
+            if pkgname not in output:
+                output[pkgname.encode('utf-8')] = {}
```
ralphbean: Why 'utf-8' here?  Would it make sense to keep it unicode while in this loop and then finally convert to byte strings at the very end?

ralphbean:
Okay, another suggestion.  This function contains two *giant* pieces on either side of ``if oformat == 'json':``.  Can you break those code-blocks into two separate functions?

```python
if oformat == 'json':
    return _do_vcs_acls_json(arguments)
else:
    return _do_vcs_acls_the_other_format(arguments)
```

Then the docstring of each of those utility methods could describe what they're
for, what the different formats are like, what uses them, etc..

pypingou:
sure can do

pkgdb2/lib/__init__.py
pypingou: I think that's a left over from when I tried to build the JSON by hand using a dict structure.

pkgdb2/lib/__init__.py
```diff
@@ -2025,6 +2025,134 @@ def bugzilla(session, name=None):
...
+
+    The output dict is something like:
+
+    {
+      pkg1: {
+        branch1: {
+          name: pkg1,
+          branch: branch1,
+          user: [user1, user2],
+          group: [group1, group2]
```
ralphbean: This looks like it should be ``groups`` not ``group`` and ``people`` not ``user`` (from reading the code below).

pypingou: No no, the key for groups is singular, see https://github.com/fedora-infra/pkgdb2/pull/166/files#diff-87909695400e927cca78825103e7f028R2139 and above :)

ralphbean: No, no.  See for ``_json`` it is plural https://github.com/fedora-infra/pkgdb2/pull/166/files#diff-87909695400e927cca78825103e7f028R2075 but for ``_text`` it is singular: https://github.com/fedora-infra/pkgdb2/pull/166/files#diff-87909695400e927cca78825103e7f028R2139

pypingou: Oh sorry, yes I was looking at the wrong method

ralphbean:
:+1:!

pypingou:
Thanks @ralphbean

PR #167, opened 2015-02-25
Monkey patch flask.jsonify to handle JSONP responses.

ralphbean:
I specifically want this for the Fedora releng-dash so it can query at
page-load and find out:

- What collections are active and which are EOL.
- What collections are "Under development" but not rawhide.. i.e., are
  we currently "branched"?

### DIFF
--- a/pkgdb2/__init__.py
+++ b/pkgdb2/__init__.py
@@ -98,7 +98,7 @@ def _monkey_patch_jsonify_jsonp():
 
         callback = flask.request.args.get('callback', None)
 
-        if callback:
+        if callback and flask.request.method == 'GET':
             if not isinstance(callback, basestring):
                 callback = callback[0]
             response.mimetype = 'application/javascript'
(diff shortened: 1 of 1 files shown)


pkgdb2/__init__.py
```diff
@@ -83,6 +83,34 @@
...
+    """
+
+    original_jsonify = flask.jsonify
+
+    def _jsonify_with_jsonp(*args, **kwargs):
+        response = original_jsonify(*args, **kwargs)
+
+        callback = flask.request.args.get('callback', None)
+
+        if callback:
```
pypingou: Could we check that the request is a GET and never a POST? Since the API is r/w I would prefer that we do not support jsonp on POST queries.

ralphbean: Sure!

ralphbean: Should be done in  b0fa924

pypingou:
I was wondering if it makes sense to have it in ``pkgdb2/__init__py`` vs having it in ``pkgdb2/api/__init__.py`` but while this is only intended for the API endpoint there is no code that restricts it to the ``/api/`` section/blueprint, so won't change much

:+1:

PR #169, opened 2015-03-03
Warn users about invalid FAS email / bugzilla account mapping

pypingou:
Currently admins are getting an email every 10 minutes informing them about invalid mappings. Then, we email the users so that they can act on it.

With this change, users will be automatically notified.

Sometime, it takes a really long time for users to react, recently we have seen more than a week. So while we do not want to spam users as much as we do for admins (every 10 minutes), we still want to pressure them into fixing it by sending them a hourly reminder.

### DIFF
--- a/utility/pkgdb-sync-bugzilla
+++ b/utility/pkgdb-sync-bugzilla
@@ -347,7 +347,7 @@ Your prompt attention in this matter is appreciated.
 
 The Fedora admins.
 """
-    data = None
+    data = {}
     if os.path.exists(DATA_CACHE):
         try:
             with open(DATA_CACHE) as stream:
@@ -364,7 +364,7 @@ The Fedora admins.
             now = datetime.datetime.utcnow()
 
             # See if we already know about this user
-            if data and user_email in data and data[user_email]['last_update']:
+            if user_email in data and data[user_email]['last_update']:
                 last_update = datetime.datetime.fromtimestamp(
                     int(data[user_email]['last_update']))
                 # Only notify users once per hour
--- a/utility/pkgdb-sync-bugzilla
+++ b/utility/pkgdb-sync-bugzilla
@@ -333,7 +333,7 @@ if __name__ == '__main__':
     for product in acls.keys():
         if product not in ('Fedora', 'Fedora EPEL'):
             continue
-        for pkg in acls[product]:
+        for pkg in sorted(acls[product]):
             if DRY_RUN:
                 print pkg
             pkgInfo = acls[product][pkg]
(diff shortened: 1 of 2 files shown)


utility/pkgdb-sync-bugzilla
```diff
@@ -293,12 +302,96 @@ def send_email(fromAddress, toAddress, subject, message):
...
     smtp = smtplib.SMTP('bastion')
     smtp.sendmail(fromAddress, toAddress, msg.as_string())
     smtp.quit()
 
 
+def notify_users(errors):
+    ''' Browse the list of errors and when we can retrieve the email
+    address, use it to notify the user about the issue.
+    '''
+    tmpl_email = """Greetings.
```
ralphbean: Make the template configurable so we can change it without a new release?

utility/pkgdb-sync-bugzilla
```diff
@@ -293,12 +302,96 @@ def send_email(fromAddress, toAddress, subject, message):
...
+https://admin.fedoraproject.org/accounts login, click on 'my account',
+then 'edit' and change your email address.
+
+If you have questions or concerns, please let us know.
+
+Your prompt attention in this matter is appreciated.
+
+The Fedora admins.
+"""
+    data = None
```
ralphbean: Why not default ``data`` to ``{}``, that way you wouldn't have to check ``if data`` all the time further down in the code.

utility/pkgdb-sync-bugzilla
```diff
@@ -293,12 +302,96 @@ def send_email(fromAddress, toAddress, subject, message):
...
+    if os.path.exists(DATA_CACHE):
+        try:
+            with open(DATA_CACHE) as stream:
+                data = json.load(stream)
+        except Exception as err:
+            print 'Could not read the json file', err
```
ralphbean: Perhaps include the value of ``DATA_CACHE`` in this error message to help debugging years down the road?

ralphbean:
I take it this gets kicked off by a cronjob every 10 minutes (and that we're going to adjust that 10 minute interval to something longer in our ansible setup?)

utility/pkgdb-sync-bugzilla
pypingou: :+1:

utility/pkgdb-sync-bugzilla
pypingou: Sure can do

utility/pkgdb-sync-bugzilla
pypingou: Good idea yes :)

pypingou:
> I take it this gets kicked off by a cronjob every 10 minutes (and that we're going to adjust that 10 
> minute interval to something longer in our ansible setup?)

No the cron will keep running every 10 minutes, so Kevin and I will still be spammed (at least for now), but users will only the notification once per hour.

ralphbean:
:+1: here

PR #170, opened 2015-03-06
Reset the cache file if no errors were found

pypingou:


ralphbean:
Can you move the string literal to the top of the file?  It's hard to read the if/else logic with that in the middle.

ralphbean:
:+1:  :)

pypingou:
Thanks for the review @ralphbean

PR #172, opened 2015-03-24
Do not automatically update branch where the user is Approved

pypingou:
This fixes the situation where the user ask for a Fedora branch (ie:
put the status to `Awaiting Review`) while he/she already has the ACL
on for example EPEL branches.
With the current behavior, all branches would be set to `Awaiting Review`
effectively leading to the user loosing the already Approved ACLs.
With this commit `Approved` ACLs are not automatically updated.

Fixes https://fedorahosted.org/pkgdb2/ticket/44


ralphbean:
:+1:

PR #173, opened 2015-03-26
Fix 'undefined' when clicking monitor toggle.

relrod:
When you click the monitoring toggle and you are not logged in, it makes
the request anyway and then appends "undefined" to the page. This looks
tacky. Instead, if we don't get a response back, just bail out and do
nothing.

Signed-off-by: Ricky Elrod <ricky@elrod.me>

### DIFF
--- a/pkgdb2/templates/package.html
+++ b/pkgdb2/templates/package.html
@@ -606,10 +606,14 @@ confusing what the difference is between that status an the per-branch status)
     $(function(){
         $("#committers").tablesorter();
 
+        {% if not(g.fas_user.username in admins
+                  or g.fas_user.username in commiters
+                  or is_admin) %}
         $('#monitor_lbl').click(function(){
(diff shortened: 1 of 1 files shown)


pypingou:
While I'm ok with the fix, I cannot reproduce the case described here. When I'm not logged in, clicking on ``Monitoring`` does not do anything for me (whether the Monitoring is on or off, no changes).

relrod:
@pypingou Updated based on IRC discussion. +1s?

pypingou:
Maybe be a little more generous/rigorous in our check? https://github.com/fedora-infra/pkgdb2/blob/fix-undefined/pkgdb2/templates/package.html#L40-L42

relrod:
@pypingou boom!

pypingou:
:+1: thanks @relrod !

PR #175, opened 2015-04-02
Add the CC email addresses to the list of person to send the email to

pypingou:


ralphbean:
Definitely need to handle the case where ``ccAddress`` is ``None``.

ralphbean:
:+1:

pypingou:
Thanks for the review and feedback @ralphbean

PR #176, opened 2015-04-07
Ensure the user is only notified once

pypingou:
When an user gets out of sync between bugzilla and FAS, we get an error
for each package that person maintains (since we cannot sync the ACLs
between pkgdb and bugzilla for each of them).
Before this commit, we would send an email to the user, once per hour,
but also once per error (thus per package that person maintains).
With this commit, we make sure to only send 1 notification per hour and
per user.


ralphbean:
:+1:

PR #183, opened 2015-04-21
Add fedmenu.

ralphbean:
For https://fedoraproject.org/wiki/Infrastructure/FY16_frontend

If you leave the fedmenu bits out of the config, it just never shows up.

### DIFF
--- a/pkgdb2/templates/master.html
+++ b/pkgdb2/templates/master.html
@@ -151,6 +151,25 @@
         });
     </script>
     {% endif %}
+
+    {% if fedmenu_url is defined %}
+    <script src="{{fedmenu_url}}/js/fedora-libravatar.js"></script>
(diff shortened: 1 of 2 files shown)


pypingou:
Looks good code wise, I'd like to test it locally before the merge though :)

pypingou:
Looks good, thanks! :+1: to merge

pypingou:
Hm, on the package page, could we make the two icons be next to each other instead of on the top of each other?

ralphbean:
> Hm, on the package page, could we make the two icons be next to each other instead of on the top of each other?

Short answer:  no.  Long answer: yes.

It currently only supports vertical alignment.. but maybe we could pass a 'alignment: "horizontal"' argument that just gets ignored for now and then someday we can figure out how to re-flow it.

Issue #179, opened 2015-04-16
Add new API listing packages retired on all active branch of Fedora/EPEL


pypingou:
These are the packages that can then be dropped from the component list on bugzilla

pypingou:
Implemented in https://github.com/fedora-infra/pkgdb2/tree/retired_pkgs

Issue #59, opened 2014-05-31
Disable retired components in bugzilla


crobinso:
This is basically a duplicate of a pkgdb RFE I filed last year:

​https://fedorahosted.org/packagedb/ticket/240

It would be nice if packagedb would disable bugzilla components when they have been retired in all current fedora releases. This is possible with the bugzilla XMLRPC API.

Basically, in pkgdb-sync-bugzilla.in, where you call server.editcomponents, you need to do

data["is_active"] = bool(component_is_retired_and_blocked_in_all_branches())

(That function doesn't exist so would need to be implemented. And that suggestion was relevant for pkgdb1 but I don't seen any instance of editcomponents in pkgdb2)

The component is then hidden in the BZ web UI component list, but bugs aren't closed, searches can still be done, etc.

An example of why this is useful: we last had a 'kvm' package in fedora 4 years ago, it was retired in favor or merging with the 'qemu' package. but the 'kvm' component exists in bugzilla, and collects misfiled bug reports all the time. yet there's no way in packagedb for me to even add myself to the bug CC list, so it's a huge pain. Not to mention dropping all the actual retired components from bugzilla will hopefully speed up filing new reports, changing component of existing bug, etc.

(originally filed for pkgdb2 at https://fedorahosted.org/pkgdb2/ticket/1 but i'd prefer to track it here)

crobinso:
Any progress on this? The Fedora 'component' list in bugzilla is still filled with long retired components

pypingou:
None other than the subject came up again recently: https://lists.fedoraproject.org/pipermail/infrastructure/2015-March/015830.html

crobinso:
This was marked fixed by https://github.com/fedora-infra/pkgdb2/issues/184 ... but I still can't tell what the expected workflow here is? Is pkgdb going to do this automatically, or is it now just easier to query to API for a list and someone still needs to manually request the components to be disabled? If the latter, that's fine, I just can't tell the current state.

Also in #184 the cover letter says:

'This should fix #59 for now as there is currently no way to call bugzilla's API to do it automatically.' 

The 'do it' here is a bit unclear, but if you mean 'disable a component' it definitely _is_ supported via bugzilla's XMLRPC interface as I mentioned in the initial comment here. I verified it is still working. However it requires 'editcomponent' bugzilla access so it won't work for tests as your regular user.

[...]

PR #184, opened 2015-04-22
Retired pkgs

pypingou:
This adds a new API endpoint listing all the packages retired in every active branch of a specified collection (defaults to: Fedora).

This should fix https://github.com/fedora-infra/pkgdb2/issues/59 for now as there is currently no way to call bugzilla's API to do it automatically.

And fixes https://github.com/fedora-infra/pkgdb2/issues/179


oddshocks:
Looks good to me :+1:

Issue #177, opened 2015-04-08
Retiring a package should turn off the monitoring flag


ralphbean:
For instance, the cacti package is [retired in rawhide](https://admin.fedoraproject.org/pkgdb/package/cacti/) but the-new-hotness just [filed](https://apps.fedoraproject.org/datagrepper/id?id=2015-3011b32f-77da-4fc9-94c3-7f83fbf51fbf&is_raw=true&size=extra-large) a [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1209663) on it (and failed to kick off a scratch build because the repo had only a ``dead.package`` file).

pypingou:
This relates a little bit to https://github.com/fedora-infra/pkgdb2/issues/59 in the way that it requires checking if the package is retired on all branches or not.

I am thinking of making an API endpoint listing the packages retired on all active branches of Fedora/EPEL. From there we could regularly update the monitoring flag and bugzilla.

pypingou:
Once #180 is done this will be easy to do

Issue #180, opened 2015-04-16
Add property to a package to know if it is retired on all active branches


pypingou:
This way we can know if the package should still be monitored or not

Relates to #177

PR #185, opened 2015-04-23
Turn off monitoring when the package is retired on all branches.

pypingou:
Fixes https://github.com/fedora-infra/pkgdb2/issues/177

### DIFF
--- a/pkgdb2/lib/__init__.py
+++ b/pkgdb2/lib/__init__.py
@@ -612,7 +612,6 @@ def update_pkg_status(session, pkg_name, pkg_branch, status, user,
             pkg = pkglisting.package
             if pkg.retired_everywhere:
                 pkg.monitor = False
-                session.add(pkg)
                 session.flush()
             if prev_status != 'Orphaned':
                 # Update Bugzilla about new owner
--- a/pkgdb2/lib/__init__.py
+++ b/pkgdb2/lib/__init__.py
@@ -608,6 +608,12 @@ def update_pkg_status(session, pkg_name, pkg_branch, status, user,
                 acl.status = 'Obsolete'
                 session.add(acl)
             session.flush()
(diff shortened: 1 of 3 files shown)


pypingou:
Also fixes https://github.com/fedora-infra/pkgdb2/issues/180

pkgdb2/lib/__init__.py
```diff
@@ -608,6 +608,12 @@ def update_pkg_status(session, pkg_name, pkg_branch, status, user,
                 acl.status = 'Obsolete'
                 session.add(acl)
             session.flush()
+            # If the package is retired everywhere, stop monitoring it
+            pkg = pkglisting.package
+            if pkg.retired_everywhere:
+                pkg.monitor = False
+                session.add(pkg)
```
ralphbean: Is this ``session.add(..)`` necessary?

pkgdb2/lib/model/__init__.py
```diff
@@ -768,7 +768,7 @@ class PackageListing(BASE):
         sa.UniqueConstraint('package_id', 'collection_id'),
     )
 
-    package = relation("Package")
+    package = relation("Package", backref=backref('listing'),)
```
ralphbean: This should probably be ``listings`` instead of ``listing`` since it is a one-to-many, not a one-to-one.

pkgdb2/lib/__init__.py
pypingou: Since I modify the package I think so, it adds it to the session as something to update in the DB.

ralphbean: I *think* that happens automatically since it should already be associated with the session from the original query.

ralphbean:
:+1: to merge, even given the question on ``session.add``.

PR #195, opened 2015-05-18
Rework the package request pages

pypingou:
* Provide a read-only page describing the request
* Rework the package request pages
* Add unit-tests


ralphbean:
:+1:

Issue #187, opened 2015-05-01
Add a new flag just for hotness scratch builds


ralphbean:
In addition to the ``monitor`` flag, it might be nice to be able to specify that "yes I want bugs filed, but I don't want auto scratch builds.

For this we would just need 1) a new column in the db for the new flag and 2) a way to toggle it.

This relates to #163 in that it will either clarify or muddle the meaning of the buttons in the UI as more are added (depending on how its done).

pypingou:
Any idea UI wise?

I thought of two ways:
* a three state button (Do not Monitor/Monitor/Monitor but no scratch build)
* a second toggle button that only appears when the first one (monitoring) is set to True

pypingou:
Or a drop-down list

Note: we could avoid adding a column to the DB by just using 3 states on the current field: Yes/No/Yes_nobuilds (or so)

ralphbean:
A three states button could be cool.. but it depends on how its executed as to whether or not it makes things more or less confusing to the user.

A drop-down list is less pretty than the fancy sliding button.. but it is probably the most clear.

Issue #163, opened 2015-02-23
The `Monitoring' flag is misleading


contyk:
The package is monitored no matter what this flag is set to; the only thing it controls, as far as I understand it [0], is whether bugs will be filed or not.

Perhaps it could be renamed to something more descriptive, e.g. `file bug reports: yes/no' or somesuch?

[0] https://fedoraproject.org/wiki/Upstream_release_monitoring#The-New-Hotness

pypingou:
Note that the title (mouse over) is a little more informative when logged in

PR #198, opened 2015-05-19
Rework the monitoring flag

pypingou:
This replace the two states flag with a three states one:
* True: monitor the package, open a bug and try a scratch build
* False: do not monitor the package
* nobuild: monitor the package, open a bug but do not try a scratch build

This should fix https://github.com/fedora-infra/pkgdb2/issues/187

As a side effect, of the new UI, I believe this will also fix https://github.com/fedora-infra/pkgdb2/issues/163


pypingou:
The new UI: 

screenshot from 2015-05-19 14 12 32

pkgdb2/static/pkgdb.css
```diff
@@ -156,6 +156,13 @@ a.wheel, a.wheel:hover, button.wheel, button.wheel:hover {
   background-position: 3px 50%;
 }
 
+label.active, label.active:hover {
+  background-image: url(https://admin.fedoraproject.org/pkgdb/static/arrow.png) !important;
```
ralphbean: Hard-coded?

ralphbean:
:+1: from me (except for that background-image url -- can that be changed?)

pkgdb2/static/pkgdb.css
pypingou: They are hard-coded for all the css, the only other option is to hard-code them as ``/pkgdb/static/...``.

We can't use flask's magic in the CSS files :(

ralphbean: I could swear you can do them relative to the css file.

ralphbean: Yeah, see [here how fedmenu does it?](https://github.com/fedora-infra/fedmenu/blob/develop/css/fedmenu.css#L70)

pypingou:
Adjusted :)

ralphbean:
:+1:  Thanks!

PR #205, opened 2015-06-26
Sometime querying anitya fails, we should be able to cope with this

pypingou:
In those case, we cannot retrieve any JSON information, so we show an
error to the user and do not fail

### DIFF
--- a/pkgdb2/ui/packages.py
+++ b/pkgdb2/ui/packages.py
@@ -349,9 +349,8 @@ def package_anitya(package, full=True):
     try:
         req = requests.get(url)
         if req.status_code != 200:
-            flask.flash(
-                'Querying anitya returned a status %s' % req.status_code,
-                'error')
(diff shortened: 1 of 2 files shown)


pkgdb2/ui/packages.py
```diff
@@ -345,8 +345,18 @@ def package_anitya(package, full=True):
         package
     )
 
-    req = requests.get(url)
-    data = req.json()
+    data = {}
+    try:
+        req = requests.get(url)
+        if req.status_code != 200:
+            flask.flash(
```
ralphbean: Would this be cleaner if you raised an exception (which would then be caught around line 357)?

ralphbean:
:+1: to the concept with one non-blocking question about style.

pkgdb2/ui/packages.py
pypingou: Could do indeed :)

PR #206, opened 2015-06-29
Koschei integration for pkgdb2

pypingou:


pypingou:
Starts to fix https://github.com/msimacek/koschei/issues/13

ralphbean:
:+1: here

pypingou:
I'll wait for the green light from the koschei folks before merging this so that when we deploy it, it actually does something :)

pypingou:
Thanks for the review though :)

mizdebsk:
Thanks @pypingou for implementing this so quickly.

In general this looks good to me. Some notes and questions:
- If I read the code correctly, anyone in `packager` group will be able to update Koschei status of any package, not just owned/maintained by them? (That behaviour meets my expectations, but I want to make sure.)
- It would be nice to allow non-packages (such as QA) to be able to edit Koschei monitoring status (`fedorabugs` looks like good candidate group), but this is not strictly required.
- packagedb-cli will need to add support for editing Koschei flag, preferably with possibility to update multiple packages at once.
- The feature at Koschei side will be implemented in such way that deployments of pkgdb2 and Koschei will not need to happen simultaneously. Koschei with the new feature will need to be deployed first - the feature will be at first disabled in config and then enabled as soon as pkgdb2 is deployed.

pypingou:
> * If I read the code correctly, anyone in packager group will be able to update Koschei status of any package, not just owned/maintained by them? (That behaviour meets my expectations, but I want to make sure.)

No only people with commit or approveacls ( and pkgdb2 admins) will be able to toggle the status:
https://github.com/fedora-infra/pkgdb2/pull/206/files#diff-87909695400e927cca78825103e7f028R2364

> *  It would be nice to allow non-packages (such as QA) to be able to edit Koschei monitoring status (fedorabugs looks like good candidate group), but this is not strictly required.

If this ends up notifying the maintainers, I would prefer to restrict this to the people that will be notified.

> * packagedb-cli will need to add support for editing Koschei flag, preferably with possibility to update multiple packages at once.

Sure can do :)

> *  The feature at Koschei side will be implemented in such way that deployments of pkgdb2 and Koschei will not need to happen simultaneously. Koschei with the new feature will need to be deployed first - the feature will be at first disabled in config and then enabled as soon as pkgdb2 is deployed.

Sounds good to me!

mizdebsk:
> No only people with commit or approveacls ( and pkgdb2 admins) will be able to toggle the status:

I'm not sure I like this solution, but maybe it will work - I'll need to give it more thought. Would this also work for group commit ACLs, such as `provenpackager` or `group::eclipse-sig`?

> If this ends up notifying the maintainers, I would prefer to restrict this to the people that will be notified.

Package maintainers will not get any notifications unless they opt-in for them in FMN app, at least for now.

mizdebsk:
> Package maintainers will not get any notifications unless they opt-in for them in FMN app, at least for now.

Except for notifications from `package.koschei.update` itself...

pypingou:
On Tue, Jun 30, 2015 at 01:52:10PM -0700, Mikolaj Izdebski wrote:
>      No only people with commit or approveacls ( and pkgdb2 admins) will be
>      able to toggle the status:
> 
>    I'm not sure I like this solution, but maybe it will work - I'll need to
>    give it more thought. Would this also work for group commit ACLs, such as
>    provenpackager or group::eclipse-sig?

provenpackager likely not, groups not sure, I would have to check

>      If this ends up notifying the maintainers, I would prefer to restrict
>      this to the people that will be notified.
> 
>    Package maintainers will not get any notifications unless they opt-in for
>    them in FMN app, at least for now.

ok.

I'd propose we try it this way for now, and if people complain we can adjust but
at least we have a consistent behavior between monitoring and koschei
integration atm.

mizdebsk:
IMO it is really important that non-owner can enable Koschei monitoring. There are several use cases that justify this. Examples:
- Suppose maintainer of `foo` wants to start using Koschei. He wants to be notified if any of his updates break packages that use `foo`, but he doesn't maintain any of that packages. It is very impractical to ask all maintainers of dependant packages to enable monitoring.
- I (and many other members of Java SIG) care about all Java packages. If there is some breakage in packages I don't own or maintain then I still want to be notified by Koschei so that I can fix them.
- As member of eclipse-sig group I am co-maintainer of most of Eclipse packages, but I have explicit ACLs only on a few of them. As co-maintainer I should be allowed to toggle Koschei monitoring flag (and upstream monitoring too).

It's understandable that only maintainers can toggle upstream monitoring status - enabling it means bugs may be filed, which could mean extra work for them. But enabling Koschei monitoring doesn't add any extra work for package maintainers, so I don't see why non-maintainers couldn't be allowed to enable monitoring.

My position is that at very minimum, at least provenpackagers should be allowed to toggle Koschei flag for any package (unless excluded by FESCo from provenpackager access) and that group ACLs shoud work.

pypingou:
>>*   packagedb-cli will need to add support for editing Koschei flag, preferably with possibility to update multiple packages at once.

> Sure can do :)

Implemented in https://github.com/fedora-infra/packagedb-cli/pull/37

ralphbean:
Is there anything left to do on this PR?  Can it be merged?

mizdebsk:
Koschei part is not implemented yet. If this PR is merged and pushed to production then it could be confusing for users - Koschei knob in pkgdb2 would have no effect.

ralphbean:
> My position is that at very minimum, at least provenpackagers should be allowed to toggle Koschei flag for any package (unless excluded by FESCo from provenpackager access) and that group ACLs shoud work.

This makes sense to me.

Furthermore, if we were to let *anyone* in the packager group toggle the koschei flag, it wouldn't necessarily guarantee unwanted notifications being sent to the maintainers.

Consider if someone unassociated with the package (a packager, but without ACLs or provenpackager status) toggles the boolean.

|   | they turn it on | they turn it off
|---|-----------------|-----------------
| maintainers want it | good. | oh no!  but, the maintainers *will* get a notification via FMN that another packager turned it off so they can deal with it.
| maintainers don't want it | not so bad.  the maintainer can leave the koschei flag on, but can disable koschei notifications in FMN | good.

pypingou:
So should we just open the switch to any packager then?

mizdebsk:
Requirements from my PoV are:
- everybody who can commit to package must be able to toggle Koschei monitoring flag, including provenpackagers and members of groups that have commit ACL (necessary condition)
- person without FAS account may not be able to toggle Koschei monitoring flag (sufficient condition)

Allowing every packager to change the flag matches these criteria, so I'm fine with it.

mizdebsk:
I think this PR can be merged now. Koschei-side code has been implemented. It is deployed in staging and ready to be tested.

pypingou:
Release time then \ó/ :)

pypingou:
Deployed: https://admin.stg.fedoraproject.org/pkgdb/package/python-pymilter/

Issue #215, opened 2015-09-02
Koschei switch-input disabled for non-commiters


mizdebsk:
Koschei monitoring switch-input is disabled for non-commiters and non-admins. All packagers are allowed to toggle monitoring (and it works, tested with API and packagedb-cli), so monitoring switch should be enabled to non-owners too.

```
                <input type="checkbox" class="switch-input" name="koschei" id="koschei"
                  {%if package.koschei %}checked="checked"{% endif %}
                  {% if not(g.fas_user.username in admins
                            or g.fas_user.username in commiters
                            or is_admin) %} disabled="disabled"{% endif %} />
```

pypingou:
You mean the action is blocked in the UI while it's allowed via the API?

mizdebsk:
> You mean the action is blocked in the UI while it's allowed via the API?

Yes, exactly.

pypingou:
Ok def a bug, thanks for the report and the pointer.

Marking it easyfix and I'll get to it if no-one fix it before me

PR #216, opened 2015-09-03
Fix koschei info in the package page

pypingou:
Any packager is allowed to change the koschei status of a package, this is true for the
API but wasn't in the UI. This is now fixed.

Fixes https://github.com/fedora-infra/pkgdb2/issues/215

The link under the ``koschei integration`` button to point to the wiki page instead of the
app directly (the wiki being clearer about what the app does).


puiterwijk:
Looks sane to me. :+1:

pypingou:
Thanks for the review

PR #222, opened 2015-09-24
fix in README

trishnaguha:


pypingou:
Oh I don't think we'll want to install pkgdb2 in the virtualenv, working in the clone is easier and this way you're sure there are no conflicts between the files you're working on and the files installed in the virtualenv.

That said python-fedora is listed in the ``requirements.txt`` which is processed below, so that entire line about install python-fedora from git could be dropped.

trishnaguha:
@pypingou I have made the change. Can you check it? 
Thank you!

pypingou:
There is a ``cd pkgdb2`` missing then, otherwise the ``pip install -r requirements.txt`` won't work :)

trishnaguha:
@pypingou  Sorry for making this a mess with so many commits.
Thanks :)

pypingou:
Looking good, thanks!

trishnaguha:
Thanks @pypingou :)

PR #225, opened 2015-09-25
Import the configuration correctly

pypingou:


ralphbean:
- Any explanation on why this change should happen?
- Also, ``.fet`` looks like a typo.  Is it supposed to be ``.get``?

pypingou:
``.fet`` is a typo indeed, the configuration of the application is ``APP``, so pkgdb2 itself can't access ``.config`` itself it needs to access it by ``.APP.config``.

puiterwijk:
Looks good to me. :+1:

PR #230, opened 2015-10-02
Be less strict with bz package review summaries.

ralphbean:
mclasen reported today that he couldn't get pkgdb to auto-populate for
this ticket:  https://bugzilla.redhat.com/show_bug.cgi?id=1229395

That's because "Review Request:" != "Review request:".

This changes the pkgdb code to check the bug component instead, and
tries to be less strict when extracting the name and description.


pypingou:
:+1: thanks!

PR #233, opened 2015-10-05
Adjust how unique an admin action can be

pypingou:
We cannot make unique an admin action based on its package, branch, user, action and status
otherwise we end up with a request closed, another open that cannot be closed with the same
status (unique constraints will fail).

What we can do otherwise is check if we have a request for that package on that branch and for this user and if so, we reset it.

This PR also adds a link to a request once it has been closed so that it can be sent to someone.

### DIFF
--- a/pkgdb2/lib/__init__.py
+++ b/pkgdb2/lib/__init__.py
@@ -1712,15 +1712,15 @@ def add_new_branch_request(session, pkg_name, clt_to, user):
     if clt_to.name == 'Fedora EPEL':
         _validate_pkg(session, clt_to.version, package.name)
 
-    action = model.AdminAction.search(
+    actions = model.AdminAction.search(
         session,
         package_id=package.id,
         collection_id=clt_to.id,
         action='request.branch',
         user=user.username,
     )
-    if action:
-        action = action.pop()
+    if actions:
+        action = actions.pop()
         action._status = status
         action.message = None
     else:
--- a/pkgdb2/lib/__init__.py
+++ b/pkgdb2/lib/__init__.py
@@ -1712,13 +1712,25 @@ def add_new_branch_request(session, pkg_name, clt_to, user):
     if clt_to.name == 'Fedora EPEL':
         _validate_pkg(session, clt_to.version, package.name)
 
-    action = model.AdminAction(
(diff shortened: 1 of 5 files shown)


pkgdb2/lib/__init__.py
```diff
@@ -1712,13 +1712,25 @@ def add_new_branch_request(session, pkg_name, clt_to, user):
     if clt_to.name == 'Fedora EPEL':
         _validate_pkg(session, clt_to.version, package.name)
 
-    action = model.AdminAction(
+    action = model.AdminAction.search(
```
ralphbean: ``.search`` returns a collection of actions?  This should be ``actions = model.AdminAction.search(...`` to make that clear.

pkgdb2/lib/__init__.py
```diff
@@ -1712,13 +1712,25 @@ def add_new_branch_request(session, pkg_name, clt_to, user):
...
+        session,
         package_id=package.id,
         collection_id=clt_to.id,
-        user=user.username,
-        _status=status,
         action='request.branch',
+        user=user.username,
     )
+    if action:
```
ralphbean: And then ``if actions:`` here.

pkgdb2/lib/__init__.py
```diff
@@ -1712,13 +1712,25 @@ def add_new_branch_request(session, pkg_name, clt_to, user):
...
+        action = action.pop()
```
ralphbean: And ``action = actions.pop()`` should do it.

pkgdb2/lib/__init__.py
pypingou: In this case it can only return 1 action as I am passing all the element that constitute the unique constraint, so it can only ever be a list of 1 item.

ralphbean: It's still a list.

I saw ``action.pop()`` and thought "wtf.  An ``AdminAction`` has a ``.pop()`` method?"

pypingou: ok

ralphbean:
Other than the cosmetic note above, :+1:.  Looks good!

Issue #229, opened 2015-10-02
change bugzilla url to bug # for request new package


nirik:
Change the form to just have a entry for the bug number instead of the entire  bugzilla url. 

Since it's always going to be https://bugzilla.redhat.com and we don't want people to use it to attack other sites.

pypingou:
They won't be able to attack other site, whatever url they enter we extract the identifier and query bugzilla with it.

That being said, we could definitely ask for only a bug number

pypingou:
The original reason for asking for the link instead of just the bug number was to be independant from bugzilla allowing:
- Other infrastructure to use pkgdb2 w/o bugzilla (or with their own bugzilla)
- Be ready to move reviews off bugzilla (fresque come here my friend!!)

PR #236, opened 2015-10-07
Improve the package request page to support giving only a ticket id

pypingou:
Fixes https://github.com/fedora-infra/pkgdb2/issues/229

### DIFF
--- a/pkgdb2/templates/package_request.html
+++ b/pkgdb2/templates/package_request.html
@@ -45,9 +45,9 @@
             $('#summary').val('');
             $('#pkgname').prop('disabled', true);
             $('#summary').prop('disabled', true);
+            $('#review_url').prop('disabled', true);
             var _bzurl = $('#review_url').val();
             var _bzid = null;
-            var _bzidint = false;
             if ($.isNumeric(_bzurl)) {
                 _bzid = _bzurl;
                 _bzidint = true;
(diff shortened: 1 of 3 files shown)


pkgdb2/templates/package_request.html
```diff
@@ -44,12 +43,20 @@
             document.body.style.cursor = 'wait';
             $('#pkgname').val('');
             $('#summary').val('');
+            $('#pkgname').prop('disabled', true);
+            $('#summary').prop('disabled', true);
             var _bzurl = $('#review_url').val();
             var _bzid = null;
-            if (_bzurl.indexOf("?id=") > -1) {
-                _bzid = _bzurl.split("?id=")[1];
+            var _bzidint = false;
```
ralphbean: It looks like you set ``_bzidint`` here, but then never use it.

pypingou: true, I had an idea but turned out to be un-necessary, will drop it

pkgdb2/templates/package_request.html
```diff
@@ -44,12 +43,20 @@
             document.body.style.cursor = 'wait';
             $('#pkgname').val('');
             $('#summary').val('');
+            $('#pkgname').prop('disabled', true);
```
ralphbean: If you're disabling pkgname and summary, should you disable/reenable the review url field too?

pypingou: disable when checking, re-enable it at the end? Good idea

pkgdb2/templates/package_request.html
```diff
@@ -44,12 +43,20 @@
...
+            $('#summary').prop('disabled', true);
+            $('#review_url').prop('disabled', true);
             var _bzurl = $('#review_url').val();
             var _bzid = null;
-            if (_bzurl.indexOf("?id=") > -1) {
-                _bzid = _bzurl.split("?id=")[1];
+            if ($.isNumeric(_bzurl)) {
+                _bzid = _bzurl;
+                _bzidint = true;
```
ralphbean: Still setting ``_bzidint`` here.

ralphbean:
OK, :+1:

pypingou:
Thanks for the review!

PR #238, opened 2015-10-08
Couple of bug fixes

pypingou:
- Make the fields readonly instead of disabling them (disable them mean they are not submitted so the form doesn't work :/)
- Ensure the review URL is a full URL not just the ticket identifier

### DIFF
--- a/pkgdb2/ui/packages.py
+++ b/pkgdb2/ui/packages.py
@@ -1236,8 +1236,7 @@ def package_request_new():
         if bz not in pkg_review_url:
             try:
                 int(pkg_review_url)
-                pkg_review_url = APP.config[
-                    'PKGDB2_BUGZILLA_URL'] + '/' + pkg_review_url
+                pkg_review_url = bz + '/' + pkg_review_url
             except (TypeError, ValueError):
                 pass
 
--- a/pkgdb2/ui/packages.py
+++ b/pkgdb2/ui/packages.py
@@ -1232,6 +1232,15 @@ def package_request_new():
         pkg_poc = flask.g.fas_user.username
         pkg_upstream_url = form.upstream_url.data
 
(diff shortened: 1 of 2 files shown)


pkgdb2/ui/packages.py
```diff
@@ -1232,6 +1232,15 @@ def package_request_new():
         pkg_poc = flask.g.fas_user.username
         pkg_upstream_url = form.upstream_url.data
 
+        bz = APP.config.get('PKGDB2_BUGZILLA_URL')
+        if bz not in pkg_review_url:
+            try:
+                int(pkg_review_url)
+                pkg_review_url = APP.config[
+                    'PKGDB2_BUGZILLA_URL'] + '/' + pkg_review_url
```
puiterwijk: You can reuse your "bz" variable here, not?

pypingou: true

pkgdb2/ui/packages.py
```diff
@@ -1232,6 +1232,15 @@ def package_request_new():
         pkg_poc = flask.g.fas_user.username
         pkg_upstream_url = form.upstream_url.data
 
+        bz = APP.config.get('PKGDB2_BUGZILLA_URL')
+        if bz not in pkg_review_url:
+            try:
+                int(pkg_review_url)
```
puiterwijk: What would happen if bugzilla_url is https://bugzilla.redhat.com/, but I enter http://bugzilla.redhat.com/123456?
The URL would become http://.../123456/https://....
(or if I made a typo in the URL).

(just a note, I think people will just copy-paste normally)

puiterwijk:
:+1:

pkgdb2/ui/packages.py
pypingou: It'll blow but won't be able to convert the url to an integer and thus will use the input as provided

PR #242, opened 2015-10-09
Add a whitelist of packager for whom we do not check their group membership

pypingou:


pkgdb2/lib/__init__.py
```diff
@@ -370,7 +372,7 @@ def set_acl_package(session, pkg_name, pkg_branch, pkg_user, acl, status,
                 'You are not allowed to approve or deny '
                 'ACLs for yourself.')
 
-    if pkg_user.startswith('group::') and acl == 'approveacls':
+    if '-' in pkg_user and acl == 'approveacls':
```
puiterwijk: I am not sure whether this is a good replacement for group:: check, since there are a lot of groups without - in their name.

puiterwijk: Okay, so there are no groups used in pkgdb2 that have no "-" in their name, but this was used to catch pseudo-users.
Good to know is that not all pseudo-users are captured by this new check: e.g. virtmaint.

@pypingou suggested to check the whitelist.

puiterwijk:
:+1: looks good to me

PR #244, opened 2015-10-10
Adding more information about PostgreSQL

farhaanbukhsh:


README.rst
```diff
@@ -66,10 +66,30 @@ pkgdb2::
 
     (my-pkgdb2-env)$ python createdb.py
 
+Setting up PostgreSQL
+=====================
+You can set up PostgreSQL instead of sqlite
+
+For setting up the PostgreSQL database you can look into the `documentation
```
pypingou: You should precise what documentation it is here: ``the `Fedora documentation about PostgresQL<http://...``

README.rst
```diff
@@ -66,10 +66,30 @@ pkgdb2::
...
+After executing all the above steps, you now need to  `Adjust Postgresql Connection Settings
+<https://github.com/fedora-infra/bodhi#3-adjust-postgresql-connection-settings>`_
+
+Now, you need to edit `/pkgdb2/default_config.py` file and replace::
+
+    DB_URL = 'sqlite:////var/tmp/pkgdb2_dev.sqlite'
+
+by::
+
+         DB_URL = 'postgresql://postgres:whatever@localhost/pkgdb2'
```
pypingou: Indentation error here :)

README.rst
```diff
@@ -66,10 +66,30 @@ pkgdb2::
 
     (my-pkgdb2-env)$ python createdb.py
 
+Setting up PostgreSQL
+=====================
+You can set up PostgreSQL instead of sqlite
```
pypingou: Maybe we could precise that a) it's not mandatory and b) why would one want to do this? :)

farhaanbukhsh: So we should give user an option kind of thing ?  I think with Postgres working with databump is a lot easire . :) So what should we put here?

pypingou: Having the data dump can make things easier, but sometime for quick testing the sqlite db can be enough.

I think we should just add a couple of sentences here explaining why one would want to use postgresql or sqlite.

pypingou:
Perfect, thanks!

farhaanbukhsh:
@pypingou all thanks to you :)

Issue #204, opened 2015-06-12
Explain more about the monitoring buttons


ralphbean:
It's not 100% clear what they do.  When you click 'bugs and builds' are you going to get bugs and builds about the package or are you going to get emails about bugs and builds?  (it's the first one)

We could add some tooltips to the buttons to explain more fully what they do, maybe?

Devyani-Divs:
Hi @ralphbean @pypingou , :) i wish to work on this bug. Can you provide me with a few links to begin with. Thankyou.

pypingou:
It would be about adding a ``title`` to this section of the code:
https://github.com/fedora-infra/pkgdb2/blob/master/pkgdb2/templates/package.html#L38-L51

farhaanbukhsh:
@ralphbean @pypingou If nobody is working on this issue I would like to take this up.

Devyani-Divs:
Hello @farhaanbukhsh , was caught up with my intern work. I will work on the bug.
@ralphbean @pypingou ,apologies for taking so long.

farhaanbukhsh:
@Devyani-Divs  cool :+1:

pypingou:
I think this has been fixed by:
https://github.com/fedora-infra/pkgdb2/pull/250
and
https://github.com/fedora-infra/pkgdb2/pull/247

PR #245, opened 2015-10-10
Fixed issue #204: Added tooltip for the buttons to explain what they do

Devyani-Divs:
Added tooltip for "Bugs and builds" button, in order to explain that they get bugs and builds about the package. @ralphbean @pypingou

### DIFF
--- a/pkgdb2/templates/package.html
+++ b/pkgdb2/templates/package.html
@@ -68,15 +68,15 @@ confusing what the difference is between that status an the per-branch status)
         <td colspan="2" align="center">
           <div id="monitoring_block">
             <input type="radio" id="no_monitoring" name="monitoring" value="0" />
-                <label for="no_monitoring" title="Click to toggle the monitor flag in the database" {%-
+                <label for="no_monitoring" title="Will toggle the monitor flag to monitor package for new releases" {%-
                     if package.monitoring_status == False %} class="active"{% endif
                     -%}>No monitoring</label> <br />
             <input type="radio" id="monitoring" name="monitoring" value="1" />
-                <label for="monitoring" title="Click to toggle the monitor flag in the database" {%-
+                <label for="monitoring" title="Will toggle the monitor flag to monitor bugs and the buildsystem for new builds" {%-
                     if package.monitoring_status == True %} class="active"{% endif
                     -%}>Bugs &amp; builds</label> <br />
             <input type="radio" id="no_build" name="monitoring" value="nobuild" />
-                <label for="no_build" title="Click to toggle the monitor flag in the database" {%-
+                <label for="no_build" title="Will toggle the monitor flag to monitor bugs for the package" {%-
                     if package.monitoring_status == "nobuild" %} class="active"{% endif
                     -%}>Bugs only</label>
             </div>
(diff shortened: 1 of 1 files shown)


pkgdb2/templates/package.html
```diff
@@ -72,7 +72,7 @@
                     if package.monitoring_status == False %} class="active"{% endif
                     -%}>No monitoring</label> <br />
             <input type="radio" id="monitoring" name="monitoring" value="1" />
-                <label for="monitoring" {%-
+                <label for="monitoring" title="Lists bugs and builds for the package" {%-
```
pypingou: Good idea but the text is wrong

Devyani-Divs:
@pypingou edited the text, does it look ok?

pypingou:
@Devyani-Divs it does not reflect what these buttons do, try to find the button in the UI, and check http://fedoraproject.org/wiki/Upstream_release_monitoring

ralphbean:
@Devyani-Divs, 01b490c comes closer, but there are three buttons there and this includes a tooltip for only one of them.  For completeness' sake, there should probably be a tooltip for each of the three explaining what each one does.

Devyani-Divs:
@ralphbean apologies ! Doing it right away !

pkgdb2/templates/package.html
```diff
@@ -68,15 +68,15 @@
         <td colspan="2" align="center">
           <div id="monitoring_block">
             <input type="radio" id="no_monitoring" name="monitoring" value="0" />
-                <label for="no_monitoring" {%-
+                <label for="no_monitoring" title="Will toggle the monitor flag to monitor package for new releases" {%-
```
pypingou: What about something more direct, like: `Turns off monitoring new releases for this package`

pkgdb2/templates/package.html
```diff
@@ -68,15 +68,15 @@
...
                     if package.monitoring_status == False %} class="active"{% endif
                     -%}>No monitoring</label> <br />
             <input type="radio" id="monitoring" name="monitoring" value="1" />
-                <label for="monitoring" {%-
+                <label for="monitoring" title="Will toggle the monitor flag to monitor bugs and the buildsystem for new builds" {%-
```
pypingou: Here maybe we could have something like: `Turns on monitoring and attempt to build new releases for this package`

pkgdb2/templates/package.html
```diff
@@ -68,15 +68,15 @@
...
                     if package.monitoring_status == True %} class="active"{% endif
                     -%}>Bugs &amp; builds</label> <br />
             <input type="radio" id="no_build" name="monitoring" value="nobuild" />
-                <label for="no_build" {%-
+                <label for="no_build" title="Will toggle the monitor flag to monitor bugs for the package" {%-
```
pypingou: And finally: `Turns on monitoring of new releases but without attempt to build it`

What do you think?

Devyani-Divs: @pypingou what can i say, apart from "just perfect" :)

pypingou:
Fixed by https://github.com/fedora-infra/pkgdb2/pull/247

PR #249, opened 2015-10-12
Fix the runserver script

pypingou:
Do the import after setting the environment variable, otherwise that
won't work


ralphbean:
:+1:

pypingou:
Thanks for the reviews :)

Issue #208, opened 2015-07-14
Nonintuitive behavior of monitoring buttons


msimacek:
The monitoring buttons are in fact radio-buttons, but since they are rendered as normal buttons, it's quite non-obvious that they work in that way. I just had to ask a colleague how to turn upstream release monitoring on, because I honestly couldn't figure it out and kept clicking on "No monitoring", expecting it to toggle. I think they should use regular radio button widget or have better graphics that convey the radio-button nature clearly. The current arrow on selected button is not enough.

jasontibbitts:
Honestly I'd suggest a dropdown instead of the radio buttons.

Devyani-Divs:
@pypingou worked on tooltip for monitoring button #204 .
Can I work on this ? Let me know, if we wish to make it a radio button or a dropdown.
Thanks :)

pypingou:
I think this got fixed by the PR above. Thanks for the report and the fix :)

PR #250, opened 2015-10-13
Fixed issue #208: Implemented default cursor for monitoring buttons when user is logged out

Devyani-Divs:
@pypingou added an "inactive" class in pkgdb.css which enforces default cursor when the fas user is not logged in, while turns into a finger-pointer for a logged in fas user or package admin/owner. Let me know if this still needs work :) Thanks.

### DIFF
--- a/pkgdb2/templates/package.html
+++ b/pkgdb2/templates/package.html
@@ -108,7 +108,6 @@ confusing what the difference is between that status an the per-branch status)
                 %} title="Only package admin or owner can turn on monitoring new releases"{%
                 else
                 %} title="Login to turn on monitoring of new releases for the package without attempting to build it"
-                class="mouse_inactive"
                 {%- endif -%}
                 {%- if package.monitoring_status == "nobuild" and g.fas_user %}
                     class="active"
--- a/pkgdb2/templates/package.html
+++ b/pkgdb2/templates/package.html
@@ -80,7 +80,7 @@ confusing what the difference is between that status an the per-branch status)
                 {%- if package.monitoring_status == False and g.fas_user %}
                     class="active"
                 {%- elif package.monitoring_status == False %}
-                    class="active inactive"
+                    class="active mouse_inactive"
                 {%- endif -%}>
                 No monitoring
               </label> <br />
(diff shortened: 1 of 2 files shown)


ralphbean:
All the ``{% if ... %}`` statements clustered in there are hard to read(!)

ralphbean:
So, as written, I think this won't work.  You'll end up with two ``class="..."`` items on the elements.

```html
<label for="no_monitoring" title="Login to turn off monitoring..." class="inactive" class="active">...
```

And I'm not sure how that will behave in the browser.  I think it has to be collapsed like this:

```html
<label for="no_monitoring" title="Login to turn off monitoring..." class="inactive active">...
```

Even then, when you read that it seems strange.  The label is both inactive **and** active?  Impossible!

It seems that the css class ``.inactive`` should probably be renamed.  How about calling it ``.disabled``?  That way the label can be both ``.disabled`` (you can't click on it) but still ``.active`` (the flag is set to ``True``).

pypingou:
Ok @Devyani-Divs so the last issue remaining is the remark made by @ralphbean, we need another name for this css class as otherwise we end-up with ``class="active inactive"``, so many something like ``mouse_inactive`` or ``default_pointer`` would be better than ``inactive``.

pkgdb2/templates/package.html
```diff
@@ -68,26 +68,55 @@
...
-                    if package.monitoring_status == "nobuild" %} class="active"{% endif
-                    -%}>Bugs only</label>
+              <label for="no_build" {%-
+                if ('packager' in g.fas_user.groups or is_admin)
+                %} title="Turns on monitoring of new releases but without attempt to build it"{%
+                elif g.fas_user
+                %} title="Only package admin or owner can turn on monitoring new releases"{%
+                else
+                %} title="Login to turn on monitoring of new releases for the package without attempting to build it"
+                class="mouse_inactive"
```
pypingou: Looks like we forgot a ``class=`` too many, this will make the element has two ``class=`` which is not allowed. So let's remove this one :)

pypingou:
:+1: for me
@ralphbean do you want to check it out as well?

pypingou:
Let's merge, thanks for your work!

Issue #192, opened 2015-05-13
[RFE] New package request does not allow to specify comaintainers


voxik:
Not sure if this is really worth of implementation, since maintainers can be added later. It is just difference from previous procedure, so I am asking if you considered this (not really issue for myself ;))

pypingou:
I thought about it and then thought it was easy enough to add co-maintainers after :)

voxik:
Ok, so may be we can close it. I have not checked documentation though. Is this mentioned somewhere?

pypingou:
Not sure about that :)

pypingou:
I'm going to mark this as an easyfix as all what is needed is to add a file to the documentation with some sort of ``frequently asked question`` page

puntogil:
if possible, how can set InitialCC with java-sig in  pkgdb for  a new package?
thanks

pypingou:
As mentioned above, you can't set a CC when making a request, but it's quite simple to add people or group afterward :)

puntogil:
with an "Package Change Request" template? i tried some time ago but dont work ...
do not accept java-sig ...
i prefer always add this for all my packages
any suggestion?
thanks

pypingou:
Once the package is created you can add other packagers, pseudo-users or groups:

1. Go to one of your package: https://admin.fedoraproject.org/pkgdb/package/apache-commons-chain/
2. Click on: Manage the <committers/watchers/admins> any of these
3. There is then an ``Add someone`` button you can use

pypingou:
Oupsy, this isn't closed yet

puntogil:
it seems that something wrong  "User "java-sig" is not in the packager group"

pypingou:
Ok so that's a missing feature as ``java-sig`` is a pseudo-user which means a) not a real account, b) hasn't signed the fpca nor a member of the packager group.

I've just find a way to go around this situation, hopefully that'll be fixed by later tonight or tomorrow.
Sorry for the inconvenience

puntogil:
thanks
see https://admin.fedoraproject.org/pkgdb/packager/java-sig/
https://fedoraproject.org/wiki/PackageDB_admin_requests#Pseudo-users_and_Groups_for_SIGs
seem reported an wrong email
should be
java-sig-commits[AT]lists.fedoraproject.org
instead of
java-devel[AT]lists.fedoraproject.org
or this is an my only mistake

pypingou:
[...]

PR #253, opened 2015-10-15
Fixed issue #192: Added the FAQ.rst file that answers this issue

Devyani-Divs:
@pypingou Added the FAQ.rst file that answers how to add comaintainers for a new package request. Please let me know if this needs work. Thanks :)


FAQ.rst
```diff
@@ -0,0 +1,18 @@
...
+
+How do I...
+===========
+
+...specify comaintainers on a new package request?
+--------------------------------------------------
+
+Once the package is created you can add other packagers, pseudo-users or groups:
+
+    1. Go to one of your package: https://admin.fedoraproject.org/pkgdb/package/apache-commons-chain/
```
pypingou: Maybe change the package name to something else ? For example ``<package name>`` ?

FAQ.rst
```diff
@@ -0,0 +1,18 @@
...
+    2. Click on: Manage the committers/watchers/admins (any of these)
```
pypingou: Maybe we could list all the buttons here since this is the official doc :)

FAQ.rst
```diff
@@ -0,0 +1,18 @@
...
+    3. There is then an Add someone button you can use.
```
pypingou: Should we put emphasis on the button, like around ``Add``

pypingou:
The file is right but I would put it withing the doc itself, see the ``doc`` folder and add it to the index.rst so that it shows up when building the doc :)

pypingou:
I was thinking of putting it in a different page, but maybe the front-page is fine

Devyani-Divs:
@pypingou different page as in, linking to FAQ.rst in the index file? Is it?

pypingou:
That was my original idea, but I'm pondering about it now, index page might be an option (especially for now, as the FAQ is short)

Devyani-Divs:
@pypingou: cool then :)

pypingou:
Let's go with this and we can always revisit later.

Thanks for your patch!

Issue #252, opened 2015-10-15
When listing requests for new packages and display page 2 I get access denied


ingvagabund:
Hi,

when I am listing all my request for new packages in pkgdb [1] and want to display page 2 (clicking next) I get:
```vim
"You are not an administrator of pkgdb"
```
This should not happen. Links for Next and Previous are incorrect. The current link for page 2 (Next) is:
```vim
https://admin.fedoraproject.org/pkgdb/admin/actions/?from_date=&page=2&package=
```
Correct link is:
```vim
https://admin.fedoraproject.org/pkgdb/packager/jchaloup/requests?page=2
```

[1] https://admin.fedoraproject.org/pkgdb/packager/jchaloup/requests

pypingou:
Thanks for the report, this is indeed a bug

PR #254, opened 2015-10-16
Fix pagination in the `My requests` page and the unit-tests

pypingou:
Fixes https://github.com/fedora-infra/pkgdb2/issues/252


ralphbean:
:+1:

Issue #257, opened 2015-10-20
Introduce an allow_retires flag on the collection table


ralphbean:
We ran into problems with the way #81 was done this week, when @tyll set the status of the F23 branch to 'Active' in order to disallow retires for final freeze.... but in the meantime, other systems have come to depend on that status field to mean other things.  We expected that it wouldn't be set to Active until after F23 GA (which is not for another few weeks).

To get around this, we were talking about adding another flag to the collection table to do what #81 wanted, so that we can then free up the 'status' field to indicate a more general status.

tyll:
another option would be a final_freeze flag to model the current development process and use it to decide whether or not retirement is possible in branch that is in the development state. This could also be used to e.g. block stable pushes in Bodhi or other tasks that depend on it - if others agree.

tyll:
btw. after this is changed, the schedule needs to be adjusted as well:
https://fedorapeople.org/groups/schedule/f-23/f-23-releng-tasks.html

pypingou:
I like the ``allow_retire`` approach more and it is self-explaining from pkgdb's point of view while ``final_freeze`` means that you must know what the final freeze is and what it implies, plus the meaning of ``final_freeze`` changes over time, just see it now, it didn't use to mean we block retirement and now it does.

PR #262, opened 2015-10-23
Add an allow_retire flag to collections

pypingou:
This flag can be used to toggle at any point in time whether someone is allowed to 
retire packages in it.

Fixes https://github.com/fedora-infra/pkgdb2/issues/257

### DIFF
--- a/alembic/versions/563dd24698a7_add_the_allow_retire_field_to_the_.py
+++ b/alembic/versions/563dd24698a7_add_the_allow_retire_field_to_the_.py
@@ -30,7 +30,7 @@ def upgrade():
                '''SET allow_retire=TRUE WHERE name='Fedora EPEL';''')
     # Set FALSE to all EOL releases - correct behavior
     op.execute('''UPDATE "Collection" '''
-               '''SET allow_retire=TRUE WHERE status='EOL';''')
+               '''SET allow_retire=FALSE WHERE status='EOL';''')
     # Set True to all releases Under Development - mostly correct, except for
     # after final freeze
     op.execute('''UPDATE "Collection" '''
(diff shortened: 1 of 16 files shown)


alembic/versions/563dd24698a7_add_the_allow_retire_field_to_the_.py
```diff
@@ -0,0 +1,48 @@
...
+            default=False,
+            server_default='False',
+            nullable=True)
+    )
+    # Set True to EPEL - correct behavior
+    op.execute('''UPDATE "Collection" '''
+               '''SET allow_retire=TRUE WHERE name='Fedora EPEL';''')
+    # Set FALSE to all EOL releases - correct behavior
+    op.execute('''UPDATE "Collection" '''
+               '''SET allow_retire=TRUE WHERE status='EOL';''')
```
ralphbean: Should this be FALSE, not TRUE?  The sql and the comment do not agree.

alembic/versions/563dd24698a7_add_the_allow_retire_field_to_the_.py
```diff
@@ -0,0 +1,48 @@
...
+    # Set True to all releases Under Development - mostly correct, except for
+    # after final freeze
+    op.execute('''UPDATE "Collection" '''
+               '''SET allow_retire=TRUE WHERE status='Under Development';''')
+
```
ralphbean: What about **Active** releases?  Should they be set to True or False?

- If True, then we're missing a sql statement here.
- If False, then perhaps add a comment to indicate that they were considered, but that the default is sufficient.

ralphbean:
:+1:

PR #264, opened 2015-10-26
Hide the Give button if the package is retired

pypingou:
This allows showing the `Un-retire` button instead which is what we want
for retired packages


pypingou:
Before:
pkgdb_old_buttons

After:
pkgdb_new_buttons

ralphbean:
Yes, :+1:

pypingou:
Thanks

Issue #260, opened 2015-10-22
clt_status option to collections api doesn't seem to work


ralphbean:
This url:  https://admin.fedoraproject.org/pkgdb/api/collections/?clt_status=Active

I think should only return collections that are active, but I see some EOL ones in there too.

pypingou:
Odd indeed

pypingou:
Ok got it and fixed it locally

PR #265, opened 2015-11-02
Fix listing collections in the API

pypingou:
Fixes https://github.com/fedora-infra/pkgdb2/issues/260


ralphbean:
:+1: here.

pypingou:
But from what I saw, nothing stands out :+1: for me

pypingou:
hm, what is this comment doing here, I thought I had put it in another PR :(

Issue #209, opened 2015-07-14
Don't request ACLs on retired branches by default


msimacek:
"Request commit access" button requests commit access on all branches. The same for "Request admin access". I think it would be more logical if it would request acls only on branches that aren't retired (user can still manage the acls per-branch if he wants to). Reasoning: There are lot of packages that have EPEL branches retired and when I want to become a comaintainer of the package, I usually don't want to unretire those.

trishnaguha:
I would like to work on it

pypingou:
@trishnaguha any progress on this? Need help with it?

trishnaguha:
@pypingou Sorry I was afk for last 5-6 days, couldn't work on the issue. I need help to make myself packager for testing purpose.

trishnaguha:
@pypingou Let me know whether I am on right track to make myself admin or not. "Request Admin access" on this package "0ad" has turned in to "Give up Admin Access". But "Request Commit Access" still shows "You must be a packager". 
![screenshot from 2015-10-24 12-42-22](https://cloud.githubusercontent.com/assets/13337039/10709460/43c20fe6-7a4d-11e5-85d6-df03c8e7fb91.png)

pypingou:
You can maybe hack https://github.com/fedora-infra/pkgdb2/blob/master/pkgdb2/ui/__init__.py#L78 to add the 'packager' group to your list of groups

trishnaguha:
@pypingou But I made the changes here https://github.com/fedora-infra/pkgdb2/blob/master/pkgdb2/lib/__init__.py#L58

pypingou:
Then you will have to adjust https://github.com/fedora-infra/pkgdb2/blob/master/pkgdb2/__init__.py#L179 as well

pypingou:
Ok so this ticket is available for anyone to work on

PR #267, opened 2015-11-03
Ignore the retired branch when requesting commit or watch* ACLs

pypingou:
Fixes https://github.com/fedora-infra/pkgdb2/issues/209

### DIFF
--- a/pkgdb2/templates/package.html
+++ b/pkgdb2/templates/package.html
@@ -68,7 +68,6 @@ confusing what the difference is between that status an the per-branch status)
         <td colspan="2" align="center">
           <div id="monitoring_block">
             <input type="radio" id="no_monitoring" name="monitoring" value="0" />
-            {{ package.monitoring_status }}
               <label for="no_monitoring" {%-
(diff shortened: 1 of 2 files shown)


pkgdb2/templates/package.html
```diff
@@ -68,7 +68,6 @@
         <td colspan="2" align="center">
           <div id="monitoring_block">
             <input type="radio" id="no_monitoring" name="monitoring" value="0" />
-            {{ package.monitoring_status }}
```
puiterwijk: This doesn't seem related to this change.
Was this a mistake, or another change you just wanted to sneak in at the same time?
(which is fine. Just want to be sure it's intended)

pypingou: It's intended, didn't think it was worth a separate PR for just this

puiterwijk: Sure, fair enough.

puiterwijk:
Looks good to me. :+1:

puiterwijk:
Still :+1:

pypingou:
Thanks! :)

PR #268, opened 2015-11-04
Fix filtering user's requests using the side panel on the web-page

pypingou:
Currently using the side panel, the url looks like:
  /packager/pingou/requests?package=&action=&status=Approved
So package and action are both ''.

With this change, they are evaluated and changed to None in those case
actually making filtering requests on the user's page working.


pypingou:
Thanks for the review

PR #271, opened 2015-11-08
Strip the input (if there is one) when a new package is requested

pypingou:
Fixes https://github.com/fedora-infra/pkgdb2/issues/272


puiterwijk:
Looks good to me.

PR #273, opened 2015-11-09
Adding link to package when action is request.package

trishnaguha:


pkgdb2/templates/list_actions.html
```diff
@@ -123,10 +123,11 @@
         </a>
       </td>
       <td>
-        {% if action.package %}
-          <a href="{{url_for('.package_info', package=action.package.name)}}">
-            {{ action.package.name }}
-          </a>
+        {% if action.package or (action.package == True and request.action == 'request.package'
```
pypingou: ``action.package`` will never be ``==True``, it cannot by definition, which means the entire second part of the condition will always fail.

pkgdb2/templates/list_actions.html
```diff
@@ -123,7 +123,8 @@
         </a>
       </td>
       <td>
-        {% if action.package %}
+        {% if action.package or (request.action == 'request.package' and
```
pypingou: What is ``request`` here? (that you call in ``request.action`` and ``request.status``)

pkgdb2/templates/list_actions.html
```diff
@@ -123,7 +123,8 @@
         </a>
       </td>
       <td>
-        {% if action.package %}
+        {% if action.package or (action.action == 'request.package' and 
+              action.status == 'Approved') %}
           <a href="{{url_for('.package_info', package=action.package.name)}}">
             {{ action.package.name }}
```
pypingou: What is the value ``action.package`` if the request is of type ``request.package``?

PR #279, opened 2015-11-23
Making sure that the request states are not accidentally changed when user is admin

### LINKED ISSUE #278: The update  admin action web form should mention that it does not change state in pkdbb
Changing a status in
https://admin.fedoraproject.org/pkgdb/admin/action/1669/status
does not process the request but only changes the state of the request. This should be stated in the web page to make sure that request states are not accidently changed.

trishnaguha:
I am not able to see the issue since I was unable to make myself admin in my local instance, let me know if this fixes #278 
Thanks :)

### DIFF
--- a/pkgdb2/templates/actions_update.html
+++ b/pkgdb2/templates/actions_update.html
@@ -41,6 +41,7 @@
     way to process requests is via the pkgdb-admin CLI tool.</span>
 </p>
 {% endif %}
+
   <table>
     <tr><td>User:</td><td>{{ admin_action.user }}</td></tr>
     <tr>
--- a/pkgdb2/templates/actions_update.html
+++ b/pkgdb2/templates/actions_update.html
@@ -35,7 +35,12 @@
 <form action="{{ url_for('.admin_action_edit_status', action_id=action_id) }}"
     method="post">
 {% endif %}
-
+{% if is_admin %}
+<p>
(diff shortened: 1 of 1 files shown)


pypingou:
I'm not sure I agree with this change, it's a process that is blocked, not a state. Don't you think?

pypingou:
I don't think this fixes the ticket #278, that ticket is asking to inform the admins that updating the status of a request does not actually process the request (ie: marking a branch request as ``approved`` does not actually create the new branch). The proper way to process requests is via the ``pkgdb-admin`` CLI tool.

pkgdb2/templates/actions_update.html
```diff
@@ -35,7 +35,12 @@
 <form action="{{ url_for('.admin_action_edit_status', action_id=action_id) }}"
     method="post">
 {% endif %}
-
+{% if is_admin %}
```
trishnaguha: @pypingou  Do I also have to check if fas user is in admin group? or only this will work?

pypingou: It is sufficient, if the variable is set/given to this template :)

ralphbean:
@pypingou, I think @trishnaguha made the update in response to your request.  Does it look okay now?

pypingou:
Yes looks good and I checked, ``is_admin`` is defined so it'll work

Issue #269, opened 2015-11-04
[RFE] Request a new package improvment


remicollet:
When using "Sync" (with from bugzilla) button to fill the "Request a new package" form, would be nice to check if

* bug have fedora-review+ flag 
* bug is assigned

This will avoid to create invalid request.

pypingou:
The code to change is in: https://github.com/fedora-infra/pkgdb2/blob/master/pkgdb2/api/extras.py#L699

PR #281, opened 2015-11-24
Improve the checks to the package review on bugzilla before letting it through

pypingou:
With these changes we check that the review was assigned to someone and
that the fedora-review flag was set to `+`.

Fixes https://github.com/fedora-infra/pkgdb2/issues/269


ralphbean:
:+1:

Issue #276, opened 2015-11-22
Some admin request cannot be updated in web UI


tyll:
The update link for some requests is blocked by the updates widget at https://admin.fedoraproject.org/pkgdb/admin/actions/

tyll:
Screenshot is here: https://till.fedorapeople.org/tmp/pkdb-unretire-problem.png

tyll:
This was noticed by @limburgher

PR #282, opened 2015-11-24
Limit the size of the content section

pypingou:
This way it does not overlap with the options box

Fixes https://github.com/fedora-infra/pkgdb2/issues/276


ralphbean:
Sure thing.  :+1:

Issue #277, opened 2015-11-22
Unretirement requests should request the similar information as new package requests


tyll:
To unretire branches, request the information that is also asked for new packages:
https://admin.fedoraproject.org/pkgdb/request/package/

- The package name can be prefilled
- The bug number is only required if the master branch needs to be unretired or if it is kept retired but an EPEL branch is unretired

pypingou:
So the information are:
* name
* bug number
* branches

For which the only info missing is the bug number (compare to the current situation), right?

pypingou:
I have worked on adding support to require a review_url, but would like to hear back from you before opening the pull-request :)

tyll:
yes, the bug number is the only important information that is missing. The other information such as summary, description and upstream URL might be worth to be updated as well. But now that I think about it, the latter information will be updated automatically by pkgdb after the package was built, so it might be omitted.

However, the bug number is also only required on certain conditions (I hope I have them right):

- if the master branch is retired for more than two weeks
- unless an EPEL branch should be unretired that is retired less than two weeks, then no re-review is needed for the EPEL branch

pypingou:
ok I can add the check, but I don't think I can for the two weeks period :(

PR #284, opened 2015-11-26
When un-retiring a package, request the review_url

pypingou:
Fixes https://github.com/fedora-infra/pkgdb2/issues/277

But this enforce the review_url regardless of the retirement date.

### DIFF
--- a/pkgdb2/forms.py
+++ b/pkgdb2/forms.py
@@ -409,3 +409,11 @@ class EditActionStatusForm(wtf.Form):
                 (status, status)
                 for status in kwargs['status']
             ]
+
+
+class UnretireForm(BranchForm):
+    """ Form to ask for a package to be un-retired. """
+    review_url = wtforms.TextField(
+        'review_url',
+        [wtforms.validators.optional()],
+    )
--- a/pkgdb2/lib/__init__.py
+++ b/pkgdb2/lib/__init__.py
@@ -1887,7 +1887,7 @@ def add_new_package_request(
     ))
 
 
-def add_unretire_request(session, pkg_name, pkg_branch, user):
+def add_unretire_request(session, pkg_name, pkg_branch, review_url, user):
     """ Register a new request to un-retire a package.
 
     This method only flushes the new objects.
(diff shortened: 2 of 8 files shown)


pkgdb2/forms.py
```diff
@@ -409,3 +409,11 @@ def __init__(self, *args, **kwargs):
                 (status, status)
                 for status in kwargs['status']
             ]
+
+
+class UnretireForm(BranchForm):
+    """ Form to ask for a package to be un-retired. """
+    review_url = wtforms.TextField(
+        'review_url',
+        [wtforms.validators.optional()],
```
puiterwijk: Would it be an idea to check for requirement of this field when generating this form, so that you can indicate whether it's required?

puiterwijk: Hmm, that would require javascript of course though... Maybe ignore this then.

pkgdb2/lib/__init__.py
```diff
@@ -1921,6 +1922,7 @@ def add_unretire_request(session, pkg_name, pkg_branch, user):
         user=user.username,
         _status='Awaiting Review',
         action='request.unretire',
```
puiterwijk: You should update the comments in api/admin.py, as that says:
    :kwarg action: restrict the actions to a specific action, options are:
        ``request.branch``, ``request.package``.

pkgdb2/ui/packages.py
```diff
@@ -788,9 +788,31 @@ def package_unretire(package, full=True):
...
+                flask.flash(
+                    'You must provide a review URL to un-retire master',
+                    'error')
+                break
+            elif br.startswith('e') and 'master' in collections and not review_url:
+                checks_ok = False
+                flask.flash(
+                    'You must provide a review URL to un-retire an EPEL '
+                    'branch if master is also retired',
+                    'error')
```
pypingou: @tyll Could we have your :+1: / :-1: on this part before merging please?

puiterwijk:
After diving through the code, it looks like this will overwrite the previously existing review_url. I am not sure whether we should, since that first review is still the basis of some branches probably.

pypingou:
Currently it's not doing anything with the review_url, just storing it in a JSON blob in the DB, there isn't an overwrite anywhere :)

puiterwijk:
Ah, okay. Maybe you should make review_url a multi-valued array? (in sql terms another table with pkgdb_id - review_url) instead of one single object, and store them all in there?

pypingou:
Something to see, atm, I prefer to keep the original review (if there is one). Allowing to link to multiple reviews will make the UI quite a bit more complex.

puiterwijk:
Fair enough, just throwing the idea out there. I don't have any further remarks that should stop this PR. :+1:

pypingou:
Ok thanks!

pypingou:
Arf, merged too quickly, I wanted to hear from @tyll first :(

tyll:
:+1:

pypingou:
ok thanks :)

Issue #286, opened 2015-11-29
not possible to set "awaiting review" for unretiring packages


ignatenkobrain:
![screenshot from 2015-11-29 17-29-07](https://cloud.githubusercontent.com/assets/2866862/11458370/d4660cb8-96be-11e5-8ba2-45b866b3aea7.png)

pypingou:
Ok, I changed the status for you so that it can be processed, but there is a bug in here somewhere :)

ignatenkobrain:
@pypingou can you do the same for https://admin.fedoraproject.org/pkgdb/package/requests/1887 ?
Thanks.

pypingou:
Fedora branch will be automatically approved once the package is un-retired. So we might actually want to close it and re-open one when master will be ready.

PR #289, opened 2015-12-01
Let the user update request to unretire to `Awaiting Review`

pypingou:
Fixes https://github.com/fedora-infra/pkgdb2/issues/286

### DIFF
--- a/pkgdb2/lib/__init__.py
+++ b/pkgdb2/lib/__init__.py
@@ -2485,7 +2485,9 @@ def edit_action_status(
             raise PkgdbException(
                 'You are not allowed to edit this request')
     elif action_status in ['Awaiting Review', 'Blocked']:
-        if not pkg_admin and not pkgdb_admin:
+        # Requester can re-set 'request.unretire' to 'Awaiting Review'
+        if (not pkg_admin and not pkgdb_admin and
+                requester and admin_action.action != 'request.unretire'):
             raise PkgdbException(
                 'You are not allowed to review this request')
     elif action_status in ['Obsolete']:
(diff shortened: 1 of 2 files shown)


pkgdb2/lib/__init__.py
```diff
@@ -2485,7 +2485,9 @@ def edit_action_status(
             raise PkgdbException(
                 'You are not allowed to edit this request')
     elif action_status in ['Awaiting Review', 'Blocked']:
-        if not pkg_admin and not pkgdb_admin:
+        # Requester can re-set 'request.unretire' to 'Awaiting Review'
+        if (not pkg_admin and not pkgdb_admin and
+                requester and admin_action.action != 'request.unretire'):
```
ralphbean: This looks potentially problematic to me.

Is it possible for ``pkgdb_admin`` to be True and ``requester`` to be false?

Does this need to look like this instead:

```python
if (not pkg_admin and not pkgdb_admin and (requester and admin_action != 'request.unretire))
```

The more I look at it, the more confused I get.  :p  (It might be okay as is).

pypingou: I know the feeling ^^

pypingou: Adjusted :)

ralphbean:
OK, after another round of thinking.  Should it be like this?

```python
if (not pkg_admin and not pkgdb_admin and not (requester and admin_action == 'request.unretire))
```

pypingou:
yeah, sounds more like it

ralphbean:
Ok.  If this is this hard to think about, maybe we should break it into two conditionals.

```python
if (not pkg_admin and not pkgdb_admin):
    raise Nope("nope")
if (requester and action != unretire):
    raise Nope("nope")
```

But that wouldn't work, right?  Because the first conditional would ```raise Nope`` before we ever got to the second one.

So, do we need an OR here?

```python
if ((not pkg_admin and not pkgdb_admin) or (requester and action != unretire)):
    raise Nope("definitely not")
```

How does that look?

pypingou:
Updated, I think this way of writing it makes it easier to read and understand

ralphbean:
:+1:

pypingou:
Thanks for the review and your patience :)

Issue #287, opened 2015-11-29
pkgdb should not allow to unretire a package in f21 after f23 was released


tyll:
it was a policy to not accept new packages in F(x-2) if F(x) was release. It seems like it got lost with the conversion to pkgdb2 for admin requests:
https://fedoraproject.org/w/index.php?title=Package_SCM_admin_requests&diff=411690&oldid=411689#New_Packages

PR #290, opened 2015-12-02
Restrict the branches proposed to un-retire to the last two Fedora releases

pypingou:
Fixes https://github.com/fedora-infra/pkgdb2/issues/287


ralphbean:
Cool, looks good here.

PR #291, opened 2015-12-03
Perform more checks around the fedora-review flag when syncing info from bugzilla

pypingou:
Fixes https://github.com/fedora-infra/packagedb-cli/issues/62


ralphbean:
:+1: here.

PR #293, opened 2015-12-04
Add namespacing to pkgdb

pypingou:
This will allow pkgdb to manage git repos for rpms but also other repos such as
docker or anything else.


pkgdb2/api/extras.py
```diff
@@ -158,18 +160,25 @@ def _vcs_acls_cache(out_format='text', eol=False, collection=None):
...
         oformat=out_format,
-        skip_pp=APP.config.get('PKGS_NOT_PROVENPACKAGER', None))
+        skip_pp=APP.config.get('PKGS_NOT_PROVENPACKAGER', None),
+        namespace=namespace)
     output = []
     if out_format == 'json':
-        output = {'packageAcls': packages,
-                  'title': 'Fedora Package Database -- VCS ACLs'}
+        output = packages
+        if 'rpms' in output:
```
ralphbean: Can we get away without special-casing this in the pkgdb server?

What reads this JSON endpoint -- the script that generates ACLs for gitolite?  I presume this change is for backwards compatibility.  By special casing this here on the pkgdb side, we'll have to *also* add special-case logic to that ACL generation script.

Can we keep the namespacing in the JSON response (to the detriment of backwards compatibility), please, and keep this stretch of code ignorant of the namespace details?

pkgdb2/api/extras.py
```diff
@@ -158,18 +160,25 @@ def _vcs_acls_cache(out_format='text', eol=False, collection=None):
...
+        output['title'] = 'Fedora Package Database -- VCS ACLs'
     else:
         for package in sorted(packages):
             for branch in sorted(packages[package]):
                 if packages[package][branch]['group']:
                     packages[package][branch]['group'] += ','
                 output.append(
-                    'avail | %(group)s%(user)s | rpms/%(name)s/%(branch)s'
+                    'avail | %(group)s%(user)s | '
+                    '%(namespace)s/%(name)s/%(branch)s'
```
ralphbean: Cool!

pkgdb2/api/packages.py
```diff
@@ -1287,9 +1291,11 @@ def api_package_request():
             cnt += 1
         collections.append(collection)
 
+    namespaces = pkgdblib.get_status(SESSION, 'namespaces')['namespaces']
```
ralphbean: Seems like this should be singular, i.e. ``namespace`` not plural, i.e. ``namespaces``.  Here, and everywhere, no?

There's never a time that a package could have multiple namespaces, correct?

pkgdb2/lib/__init__.py
```diff
@@ -2579,3 +2602,88 @@ def get_retired_packages(session, collection):
...
+        return 'Namespace "%s" created' % namespace
+    except SQLAlchemyError, err:  # pragma: no cover
+        pkgdb2.LOG.exception(err)
+        session.rollback()
+        raise PkgdbException(
+            'Could not add Namespace "%s" to the database.' % namespace)
+
+
+def drop_namespace(session, namespace, user):
+    """ Remove a namespace to the database.
```
ralphbean: ``s/to the database/from the database/``

pkgdb2/templates/list_packages.html
```diff
@@ -33,7 +33,7 @@
 
 {% block content %}
 
-<h1>Search packages</h1>
+<h1>Search packages{% if namespace %} ({{ namespace }}) {% endif %}</h1>
```
ralphbean: Nice!

pkgdb2/templates/master.html
```diff
@@ -47,8 +47,13 @@
               <option value="retired"
               {% if select=='retired' %} selected {% endif %}>
                   Retired</option>
+              {% for ns in namespaces %}
+              <option value="{{ ns }}"
+              {% if select==ns %} selected {% endif %}>
```
ralphbean: Should this be ``if select=='rpms'`` ?  The default?

ralphbean: Or, no... I'm confused by this block, I guess.  Something seems off.

pkgdb2/templates/package.html
```diff
@@ -200,8 +200,13 @@
 {% block content %}
 
 <section id="pkg_info">
-  <h1 class="inline" property="doap:name">{{ package.name }}</h1>{% if package.upstream_url
-      %} (<a class="inline" href="{{ package.upstream_url }}">upstream</a>){% endif %}
+  <h1 class="inline">
+      <span class="italic">{{ package.namespace | capitalize}}:</span>
```
ralphbean: Replace the ``:`` with ``/``?  We want to encourage people to think of the namespaces as directories everywhere I think (that was what I got out of the devel list discussion).  I.e.:

- rpms/nethack
- rpms/cockpit

and

- dockerfiles/cockpit

In this section of the template, can we ensure there is no space inbetween the ``/`` and the text on either side, too?

pkgdb2/api/packages.py
ralphbean: I see now -- this is getting all of the namespaces in order to populate a drop-down menu, correct?  If that is correct, then this plural usage is also correct.  (my bad).

pkgdb2/api/extras.py
pypingou: It was for backward compatibility but I'm fine with breaking it. It's an ``extra`` API endpoint anyway and we haven't promise to keep it stable.

pkgdb2/api/packages.py
pypingou: It's the idea yes (getting all of them from the DB to populate the drop-down list)

pkgdb2/templates/master.html
pypingou: select is provided to the template and here, in the options for searching pkgdb, we try to make sure that if you searched for ``packagers`` or ``docker``, the drop-down of the search options is set to what you searched (``packagers`` or ``docker``)

pkgdb2/templates/package.html
pypingou: ok with the ``:`` -> ``/`` move, I'll check for the space.

pkgdb2/templates/master.html
ralphbean: Cool, I understand now.

pypingou:
Tests are broken currently, will fix them shortly

ralphbean:
Cool.  This looks good.

Aside from fixing the VCS regeneration script to handle the new output.. can you think of any other systems that might need to be modified to handle this stuff?

pypingou:
No that's the only one I can think of right now

pypingou:
Rebased and adjusted :)

ralphbean:
OK, :+1: here (assuming we can adjust the gitolite ACL script at release/deploy time).

ralphbean:
Time to merge, no?  :+1: here.  :)

pypingou:
First thing on the list for tomorrow, release 2.0 :)

PR #294, opened 2015-12-07
Fix the API to actually return the ACLs when requesting them

pypingou:


pkgdb2/api/packages.py
```diff
@@ -905,22 +905,29 @@ def api_package_list(pattern=None):
         critpath = False
     elif critpath:
         critpath = True
-    acls = bool(flask.request.args.get('acls', False))
+    acls = flask.request.args.get('acls', False)
+    if str(orphaned).lower() in ['0', 'false']:
```
ralphbean: Do you mean ``acls`` instead of ``orphaned`` here?

pypingou: oups :]

ralphbean: for tech-debt-fighting week, one task could be to convert all little blocks like this to use an ``asbool`` function or something like that.

```python
acls = asbool(flask.request.args.get('acls', False))

def asbool(value):
    if value is None:
        return False
    if isinstance(value, bool):
        return value
    return str(value).lower() in ['1', 'true', 'yes', 'please']  # etc...

ralphbean:
:+1:

pkgdb2/api/packages.py
pypingou: :+1: for me

pypingou: One way around this is to create a Null object and to use it:

````python
class Null(Object):
   pass

branch = asbool(flask.request.args.get('branch', Null))

def asbool(value):
    if value is Null:
        return None
    if isinstance(value, bool):
        return value
    return str(value).lower() in ['1', 'true', 'yes', 'please']
````

pypingou:
Thanks

PR #295, opened 2015-12-09
More namespacing adjustments

pypingou:
This allows to the namespace to be specified in the URLs, adjust the templates and unit-tests accordingly


pkgdb2/api/packages.py
```diff
@@ -194,17 +199,21 @@ def api_package_edit():
     output = {}
 
     pkg_status = pkgdblib.get_status(SESSION, 'pkg_status')['pkg_status']
```
ralphbean: It's not directly related to this change, but it pops out at me seeing ``namespaces`` next to it.  This should probably be ``pkg_statuses`` to indicate that it is plural, a collection, a list.

pkgdb2/api/packages.py
```diff
@@ -401,7 +414,9 @@ def api_package_unorphan():
 
     pkgnames = flask.request.form.getlist('pkgnames', None)
     branches = flask.request.form.getlist('branches', None)
+    namespace = flask.request.form.get('namespace', 'rpms')
     poc = flask.request.form.get('poc', None)
+    namespace = flask.request.form.get('namespace', None)
```
ralphbean: This looks like a duplicate line that can be removed.

pkgdb2/api/packages.py
```diff
@@ -639,7 +660,9 @@ def api_package_unretire():
 @API.route('/package')
 @API.route('/package/<pkgname>/')
 @API.route('/package/<pkgname>')
-def api_package_info(pkgname=None):
+@API.route('/package/<namespace>/<pkgname>/')
+@API.route('/package/<namespace>/<pkgname>')
+def api_package_info(namespace='rpms', pkgname=None):
```
ralphbean: Wow!  Does defaulting to ``'rpms'`` here really work if you visit ``/package/nethack``?

pkgdb2/api/packages.py
pypingou: agreed, but maybe for another PR :)

pkgdb2/api/packages.py
```diff
@@ -781,7 +806,9 @@ def api_package_info(pkgname=None):
 @API.route('/packages')
 @API.route('/packages/<pattern>/')
 @API.route('/packages/<pattern>')
-def api_package_list(pattern=None):
+@API.route('/package/<namespace>/<pattern>/')
+@API.route('/package/<namespace>/<pattern>')
+def api_package_list(namespace=None, pattern=None):
```
ralphbean: Should the default value for ``namespace`` be ``'rpms'`` here in the function signature?  That would maintain symmetry with the ``api_package_info`` function above.

pkgdb2/api/packages.py
pypingou: It's not enough no, but it ensure there is always a namespace provided.

pkgdb2/api/packages.py
pypingou: gone

pkgdb2/api/packages.py
```diff
@@ -1061,13 +1094,14 @@ def api_package_critpath():
     elif str(critpath).lower() in ['0', 'false']:
         critpath = False
 
-    if pkgnames and branches:
+    if pkgnames and branches and namespace:
```
ralphbean: Below, you write:

```python
         if not pkgnames:
             detail.append('pkgnames: This field is required.')
         if not branches:
             detail.append('branches: This field is required.')
```

If ``namespace`` is going to be required here, it should be added to that error handling also.

pkgdb2/api/packages.py
```diff
@@ -1105,9 +1139,9 @@ def api_package_critpath():
     return jsonout
 
 
-@API.route('/package/<package>/monitor/<status>', methods=['POST'])
+@API.route('/package/<namespace>/<package>/monitor/<status>', methods=['POST'])
```
ralphbean: No default fallback here?

pkgdb2/api/packages.py
pypingou: I actually wonder if the approach used here isn't better (avoids ``''`` iirc)

pkgdb2/api/packages.py
```diff
@@ -1168,9 +1203,9 @@ def api_monitor_package(package, status):
     return jsonout
 
 
-@API.route('/package/<package>/koschei/<status>', methods=['POST'])
+@API.route('/package/<namespace>/<package>/koschei/<status>', methods=['POST'])
 @packager_login_required
-def api_koschei_package(package, status):
+def api_koschei_package(namespace, package, status):
```
ralphbean: Same here, no default fallback?

pkgdb2/api/packages.py
```diff
@@ -1370,8 +1406,8 @@ def api_package_request():
     return jsonout
 
 
-@API.route('/request/branch/<package>', methods=['POST'])
-def api_branch_request(package):
+@API.route('/request/branch/<namespace>/<package>', methods=['POST'])
+def api_branch_request(namespace, package):
```
ralphbean: Same here, no default fallback?

pkgdb2/lib/__init__.py
```diff
@@ -298,7 +302,7 @@ def get_acl_package(session, pkg_name, pkg_clt=None, eol=False):
         found in the database with the name ``pkg_name``.
 
     """
-    package = model.Package.by_name(session, pkg_name)
+    package = model.Package.by_name(session, pkg_name, namespace)
```
ralphbean: In some functions, the arguments are ``(session, namespace, pkg_name, ....)`` and here it is ``(session, pkg_name, namespace)``.  It would help for long term maintainability and debugging if it was consistent.  I like the idea of having ``namespace`` come before ``pkg_name``.  Can we standardize on that?

pkgdb2/lib/__init__.py
```diff
@@ -931,13 +941,13 @@ def search_actions(
         count=count)
 
 
-def search_logs(session, package=None, packager=None,
-                from_date=None, page=None,
-                limit=None, count=False):
+def search_logs(session, package=None, namespace=None, packager=None,
```
ralphbean: Here, also, namespace comes *after* ``package`` in the signature.. but having it come first seems like it would be more consistent with other functions, imho.

pkgdb2/lib/__init__.py
```diff
@@ -1259,9 +1270,10 @@ def edit_collection(session, collection, clt_name=None, clt_version=None,
             raise PkgdbException('Could not edit Collection.')
 
 
-def edit_package(session, package, pkg_name=None, pkg_summary=None,
-                 pkg_description=None, pkg_review_url=None,
-                 pkg_upstream_url=None, pkg_status=None, user=None):
+def edit_package(
+        session, package, pkg_name=None, pkg_summary=None,
+        pkg_description=None, pkg_review_url=None, pkg_upstream_url=None,
+        pkg_status=None, user=None):
```
ralphbean: Namespace is notably absent here.. but I guess that makes sense.  When we would edit the namespace of a package entry?  Never.

pkgdb2/lib/model/__init__.py
```diff
@@ -461,7 +461,7 @@ def get_acl_packager(
         return query.all()
 
     @classmethod
-    def get_acl_package(cls, session, user, package,
+    def get_acl_package(cls, session, user, package, namespace,
```
ralphbean: Perhaps change the order here, as mentioned in other places.

pkgdb2/ui/acls.py
```diff
@@ -39,14 +39,16 @@
 # pylint: disable=E1101
 
 
-@UI.route('/acl/<package>/request/', methods=('GET', 'POST'))
+@UI.route('/acl/<namespace>/<package>/request/', methods=('GET', 'POST'))
 @fas_login_required
-def request_acl(package):
+def request_acl(namespace, package):
```
ralphbean: Can we do a default ``namespace=rpms`` here like in that one other flask endpoint?

pkgdb2/api/packages.py
ralphbean: But.. if it doesn't work when you visit ``/package/nethack``, how would that default value ever be provided?

pkgdb2/ui/packages.py
```diff
@@ -44,6 +44,7 @@
 
 @UI.route('/packages/')
 @UI.route('/packages/<motif>/')
+@UI.route('/packages/<namespace>/<motif>/')
 def list_packages(motif=None, orphaned=None, status=None, namespace=None,
```
ralphbean: Does ``namespace`` need to be added to the signature here?

pkgdb2/ui/packages.py
```diff
@@ -142,15 +146,18 @@ def list_retired(motif=None):
 # pylint: disable=R0914
 ## Too many statements
 # pylint: disable=R0915
-@UI.route('/package/<package>/')
-def package_info(package):
+#@UI.route('/package/<package>/')
+@UI.route('/package/<namespace>/<package>/')
+def package_info(package, namespace='rpms'):
```
ralphbean: Signature is backwards here?

pkgdb2/ui/packages.py
```diff
@@ -260,7 +267,8 @@ def package_info(package):
 
 
 @UI.route('/package/<package>/timeline')
-def package_timeline(package):
+@UI.route('/package/<namespace>/<package>/timeline')
+def package_timeline(package, namespace='rpms'):
```
ralphbean: Same here.  namespace first?

pkgdb2/ui/packages.py
```diff
@@ -327,9 +337,11 @@ def package_timeline(package):
...
 
 
-@UI.route('/package/<package>/anitya')
-@UI.route('/package/<package>/anitya/<full>')
-def package_anitya(package, full=True):
+#@UI.route('/package/<package>/anitya')
+#@UI.route('/package/<package>/anitya/<full>')
+@UI.route('/package/<namespace>/<package>/anitya')
+@UI.route('/package/<namespace>/<package>/anitya/<full>')
+def package_anitya(package, namespace='rpms', full=True):
```
ralphbean: Same here.  namespace first?

pkgdb2/api/packages.py
pypingou: I'll drop it indeed

pkgdb2/api/packages.py
pypingou: agreed

pkgdb2/lib/__init__.py
pypingou: ok for me

pkgdb2/lib/__init__.py
pypingou: It's not needed here because we have the package object (rather than just its name), so the namespace is already attached and indeed, we cannot/don't want to allow editing it.

pypingou:
Ok, I'll be dropping all the ``namespace='rpms'`` from this PR, the idea here is to break all things to what we want pkgdb to be, I have another PR after this one that brings back backward compatibility for the URLs of the UI (GET only).

pkgdb2/api/packages.py
pypingou: Nope, this is a POST only endpoint, I'm in favor of breaking the API there.

What do you think?

pkgdb2/api/packages.py
pypingou: Same as above

pkgdb2/api/packages.py
ralphbean: Sure.  I guess I thought it was possible with two ``@API.route(..`` decorators, but if not, then let's go for it.

pkgdb2/lib/__init__.py
ralphbean: Agreed.  Cool.  :)

ralphbean:
Hm.  Can we add an ``api_version`` endpoint that returns '2.0'?

Then, we can modify pkgdb2client to check that url first, and use the new urls if it is 2.0 and use the old urls if it gives a ``404`` error.

pypingou:
> Hm. Can we add an api_version endpoint that returns '2.0'?

You mean, like: https://admin.fedoraproject.org/pkgdb/api/version ? :)

ralphbean:
Yes, excellent.

So, what version should we check for to see if this stuff is live?

pypingou:
It's definitely going to be a ``2.x``

ralphbean:
:+1: to merge.

pypingou:
Let's do it!

PR #297, opened 2015-12-10
Fix the bugzilla integration when requesting a new package

pypingou:


puiterwijk:
Looks sane to me.

pypingou:
Thanks

PR #298, opened 2015-12-11
Fix processing new package requests

pypingou:
This commit does two fixes:
- It uses the proper key to retrieve the name of the package in the JSON
blob stored in the database to represent an admin action request
- If no package could be found with the provided package name, skip
without failing, this will allow pkgdb-admin to process the whole queue


puiterwijk:
Ack

pypingou:
Thanks, I'll cut a new release with this :)

PR #302, opened 2015-12-16
Fix the page listing all the packages

pypingou:
- Add the namespace to the link as everywhere else
- Allow filtering the packages for a given namespace
- Fix the runserver script (unrelated to the other changes but still needed)


puiterwijk:
Ack

PR #304, opened 2015-12-18
Fix giving ACLs to someone

pypingou:
To give ACLs to someone we need the namespace (to figure out which
package we're talking about), but this namespace might not actually be
passed on by the form, in this case, we'll just use the one provided in
the URL.

Adjust the unit-tests to cover this situation

### DIFF
--- a/pkgdb2/ui/acls.py
+++ b/pkgdb2/ui/acls.py
@@ -262,6 +262,9 @@ def package_give_acls(namespace, package):
         namespaces=acls['namespaces'],
     )
     form.pkgname.data = package
+    if str(form.namespace.data) in ['None', '']:
+        form.namespace.data = 'rpms'
+
     if form.validate_on_submit():
         pkg_branchs = form.branches.data
         pkg_acls = form.acl.data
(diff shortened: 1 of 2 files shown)


pkgdb2/ui/acls.py
```diff
@@ -262,6 +262,9 @@ def package_give_acls(namespace, package):
         namespaces=acls['namespaces'],
     )
     form.pkgname.data = package
+    if str(form.namespace.data) in ['None', '']:
```
ralphbean: Why not ``if not form.namespace.data:``?

ralphbean:
:+1:.

pkgdb2/ui/acls.py
pypingou: I thought there was a reason but can't remember it right now, so sure :)

pypingou: And the reason is that it doesn't work, I suspect this is something to do with wtforms.

Checking further into it:
``
print form.namespace.data
print type(form.namespace.data)
``
Returns
``
None
<type 'unicode'>
``
That's why the ``if not`` fails :-/

ralphbean: OH!  Hm.  Ok.

PR #305, opened 2015-12-18
Handle the situation where there is no package linked to the request

pypingou:
Adjust unit-tests to cover this situation


ralphbean:
:+1:

PR #312, opened 2016-01-05
Fix user-specific fedmenu in package timeline view

mizdebsk:
When no user is specified for filtering, timeline view displays user-specific fedmenu with broken links.

Reproducer:
- go to https://admin.fedoraproject.org/pkgdb/package/ant/timeline

Actual result:
- user-specific fedmenu is displayed
- user-specific fedmenu contains broken links

Expected result:
- user-specific fedmenu is not displayed

### DIFF
--- a/pkgdb2/templates/master.html
+++ b/pkgdb2/templates/master.html
@@ -169,7 +169,11 @@
             'user': '{{packager}}',
           {% endif %}
           {% if package is defined %}
+           {% if package.name is defined %}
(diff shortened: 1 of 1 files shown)


mizdebsk:
Package-specific fedmenu also displays incorrect links and this PR attempts to fix it too.

pypingou:
Tested locally, works fine.

Thanks for finding and fixing this! :)

Issue #319, opened 2016-02-01
namespace is not included in dead.package git urls


apevec:
namespace parameter was added in https://github.com/fedora-infra/pkgdb2/commit/08104fc65744593b08a3d1a48c18c44b3016c9f4
but def api_dead_package(pkg_name, clt_name)[1] doesn't use it and clicking "Retired" links fail with No `dead.package` file found in javascript popup.

[1] https://github.com/fedora-infra/pkgdb2/blob/master/pkgdb2/api/extras.py#L628

pypingou:
Which package were you checking?

Because right now, since there is only one namespace in prod (``rpms``) and cgit should be backward compatible, this should have no impact.

apevec:
e.g. https://admin.fedoraproject.org/pkgdb/package/rpms/openstack-nova/ but also others

Generated javascript has:
var url = "/pkgdb/api/dead/package/openstack-nova/?namespace=rpms";

pypingou:
Looks like you found the bug, the correct url is: https://admin.fedoraproject.org/pkgdb/api/dead/package/openstack-nova/master?namespace=rpms

Is this something you would be interesting in fixing ?

apevec:
So I had a look and it's unholy mix of flask and javascript!
PR coming, tested locally w/ production dump.

pypingou:
> So I had a look and it's unholy mix of flask and javascript!

As in, you think we could improve on this or is this just an judgment call?

apevec:
No, just general bitching, ignore me. I'm not frontend developer and I just don't like javascript.

PR #320, opened 2016-02-02
Include namespace in api_dead_package

apevec:
Fixes https://github.com/fedora-infra/pkgdb2/issues/319

### DIFF
--- a/pkgdb2/api/extras.py
+++ b/pkgdb2/api/extras.py
@@ -625,7 +625,8 @@ def api_koschei():
 
 
 @API.route('/dead/package/<pkg_name>/<clt_name>')
-def api_dead_package(pkg_name, clt_name):
+@API.route('/dead/package/<namespace>/<pkg_name>/<clt_name>')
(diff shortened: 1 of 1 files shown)


pkgdb2/api/extras.py
```diff
@@ -625,7 +625,8 @@ def api_koschei():
 
 
 @API.route('/dead/package/<pkg_name>/<clt_name>')
-def api_dead_package(pkg_name, clt_name):
+@API.route('/dead/package/<namespace>/<pkg_name>/<clt_name>')
```
pypingou: This won't be backward compatible, we need either to support the old route or to add a second one

apevec: This is exactly what I did here, added a second route.

pypingou: Oh, stupid me, sorry about that

pypingou:
Tested locally as well, works great, thanks a lot for catching and fixing this :)

PR #322, opened 2016-02-11
Update update_package_info.py to support namespacing and report all unknown packages

pypingou:


ralphbean:
:+1:, nice catch!

pypingou:
Thanks

Issue #324, opened 2016-03-01
Packages search results next links not working


swt2c:
If I search packages in pkgdb for something like `python*`, the 'Next' link doesn't include my search.  In other words, if I click it, it results in page 2 of **all** packages, not my search.

PR #325, opened 2016-03-01
Set correctly the motif in the template so that pagination works in search

pypingou:
Fixes https://github.com/fedora-infra/pkgdb2/issues/324


ralphbean:
Nice and simple.  :+1:  :)

pypingou:
Thanks

PR #327, opened 2016-03-02
Allow specifying co-maintainers when requesting a new package to be added

pypingou:
Fixes https://fedorahosted.org/pkgdb2/ticket/54


puiterwijk:
I think you should make sure during the processing of the request that the co-maintainers are also packagers, but :+1: on this request patch.

pypingou:
Double checked, we're good to go :)

PR #331, opened 2016-03-16
Optimize the pkgdb2_branch script used when branching

pypingou:
Instead of iterating through the element in the DB and doing one query at a time,
we do two queries:
1/ creates the branch for all the packages
2/ sync the ACLs so that the new branch has the same has rawhide.

With this change, doing a branching 
* on my local machine from 1h45 to 25 minutes
* in stg from 5h+ to 32minutes

### DIFF
--- a/pkgdb2/lib/__init__.py
+++ b/pkgdb2/lib/__init__.py
@@ -28,6 +28,7 @@ import json
 
 import sqlalchemy
 
+from datetime import datetime
 from datetime import timedelta
 from sqlalchemy.orm import sessionmaker
 from sqlalchemy.orm import scoped_session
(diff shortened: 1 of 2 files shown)


ralphbean:
Nice!  A side effect of this is that we'll need to run the genacls.sh script on pkgs01 afterwards, right? It is usually triggered by a fedmsg message.. but this won't create those messages anymore (if I'm reading it right).

ralphbean:
:+1: here

pypingou:
> A side effect of this is that we'll need to run the genacls.sh script on pkgs01 afterwards, right? 

Likely but we will probably want to run it manually anyway since otherwise branches are created at the first commit in the repo instead of the last commit in the master branch.
(The SOP is updated, on the wiki which is the out-dated location :s)

ralphbean:
Cool, :+1: still.

PR #333, opened 2016-04-11
Add the namespace in the JSON representation of a package

pypingou:


pypingou:
I'm confused, does that mean you agree with the change? We normally comment about the review and let the person who opened the PR merge if they have the rights to do so, otherwise, comment and merge, but w/o comments it's confusing

tyll:
yes, I agree with the change. I am about to update the packagedb-cli PR and wanted to mention it there.

PR #336, opened 2016-04-13
Monitoring status

tyll:
Remove the old, no longer needed set_monitoring_status script and use %license in the spec file


pypingou:
Indeed this is no longer needed, thanks

tyll:
thank you

PR #337, opened 2016-04-14
Fix the update_package_info script

pypingou:


pypingou:
This is already fixed in production

tyll:
looks good to me

pypingou:
Thanks for the review

PR #340, opened 2016-04-15
Check review_url upon asking to un-retire a package

pypingou:


pkgdb2/ui/packages.py
```diff
@@ -905,6 +920,7 @@ def package_unretire(namespace, package, full=True):
                         flask.flash(str(err), 'error')
                         SESSION.rollback()
                     except SQLAlchemyError, err:
+                        print err
```
puiterwijk: Is this a leftover from debugging? Perhaps we should instead use the logging functions.

pkgdb2/ui/packages.py
```diff
@@ -878,6 +880,19 @@ def package_unretire(namespace, package, full=True):
...
                 break
 
+        # Check the review URL
+        bz = APP.config.get('PKGDB2_BUGZILLA_URL')
+        try:
+            int(review_url)
+            review_url = bz + '/' + review_url
+        except (TypeError, ValueError):
+            pass
+        if review_url and bz not in review_url:
```
puiterwijk: Perhaps it's an idea to make sure that it also ends in the bug ID?
So that people aren't going to submit URLs like https://bugzilla.redhat.com/show_bug.cgi?id=123456#c2 or something.

puiterwijk: Also, we might be able to support both the /show_bug.cgi?id=123456 and /123456 urls, and always display the /123456?

pkgdb2/ui/packages.py
pypingou: most like remain yes

pkgdb2/ui/packages.py
pypingou: Regarding showing, it's in a different part of the app since this is the review to unretire not the original review we link to in the package view, but I could probably use the same logic here if we want.

For the comment, we could yes. This makes me think we probably need an utility method to check the BZ url in general.

pypingou:
Ok adjusted :)

puiterwijk:
Looks sane to me :+1:

pypingou:
Thanks

PR #347, opened 2016-04-30
Improve the mass-branching script

pypingou:
With some user interaction it can now update the rawhide collection
and then create the new branch before doing the branching.
Meaning there is no longer any need to use the web-ui, then go CLI
to do the branching.


lmacken:
Looks good! :+1:

PR #355, opened 2016-05-26
Check namespace policy on branch request.

### LINKED ISSUE #352: Namespace policy doesn't cover new package request
@ralphbean when testing things on stg, I could request fine a new package to be added to the `modules` namespace with more branches than just the `master` one

ralphbean:
Added a test for it as well.

Fixes #352.


pypingou:
:+1: to this I wonder if we could add a little JS to the HTML to reduce the list of branches available as option in the UI to make it friendlier

ralphbean:
> I wonder if we could add a little JS to the HTML to reduce the list of branches available as option in the UI to make it friendlier

I'll file a separate issue for it and assign it to myself.

PR #359, opened 2016-06-03
Do not hard-code packager and provenpackager in the code

pypingou:
Instead, allow them to be set in the configuration file


ralphbean:
Looks good here.  :+1:

pypingou:
Thanks :)

PR #360, opened 2016-06-04
Properly handle bugzilla urls for unretirement requests

tyll:
This PR first fixes the test suite to not report errors because of undefined exceptions, adds a test for the function to identify the probem, fixes the problem and also fixes some PEP8 violations in the touched files.

This should fix the problem in  https://fedorahosted.org/rel-eng/ticket/6426

### DIFF
--- a/pkgdb2/api/acls.py
+++ b/pkgdb2/api/acls.py
@@ -34,7 +34,7 @@ import pkgdb2.forms as forms
 import pkgdb2.lib as pkgdblib
 from pkgdb2 import SESSION, APP
 from pkgdb2.api import API
-
+from pkgdb2.lib.exceptions import PkgdbBugzillaException, PkgdbException
 
 ## Some of the object we use here have inherited methods which apparently
 ## pylint does not detect.
@@ -143,7 +143,7 @@ def api_acl_update():
             SESSION.commit()
             output['output'] = 'ok'
             output['messages'] = messages
-        except pkgdblib.PkgdbException, err:
+        except PkgdbException as err:
             SESSION.rollback()
             output['output'] = 'notok'
             output['error'] = str(err)
@@ -234,11 +234,11 @@ def api_acl_reassign():
                 )
                 SESSION.commit()
                 messages.append(message)
-            except pkgdblib.PkgdbBugzillaException, err:  # pragma: no cover
+            except PkgdbBugzillaException as err:  # pragma: no cover
                 APP.logger.exception(err)
                 SESSION.rollback()
                 errors.add(str(err))
-            except pkgdblib.PkgdbException, err:
+            except PkgdbException as err:
                 SESSION.rollback()
                 errors.add(str(err))
 
--- a/pkgdb2/api/admin.py
+++ b/pkgdb2/api/admin.py
@@ -151,7 +151,7 @@ List admin actions
             status=status,
             count=True,
         )
-    except pkgdblib.PkgdbException, err:  # pragma: no cover
+    except pkgdblib.PkgdbException as err:  # pragma: no cover
         SESSION.rollback()
         output['output'] = 'notok'
         output['error'] = str(err)
@@ -336,7 +336,7 @@ Edit Admin Action status update
                 SESSION.commit()
                 output['output'] = 'ok'
                 output['messages'] = [message]
-            except pkgdblib.PkgdbException, err:  # pragma: no cover
+            except pkgdblib.PkgdbException as err:  # pragma: no cover
                 # We can only reach here in two cases:
                 # 1) the user is not an admin, but that's taken care of
                 #    by the decorator
--- a/pkgdb2/api/admin.py
+++ b/pkgdb2/api/admin.py
@@ -31,6 +31,7 @@ import pkgdb2.lib as pkgdblib
 import pkgdb2.forms
 from pkgdb2 import SESSION, is_admin
 from pkgdb2.api import API, get_limit
+from pkgdb2.lib.exceptions import PkgdbException
 
 
 @API.route('/admin/actions/')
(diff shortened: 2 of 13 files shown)


pkgdb2/lib/__init__.py
```diff
@@ -1972,15 +1975,21 @@ def add_new_package_request(
     if pkg_collection.startswith(('el', 'epel')):
         _validate_pkg(session, pkg_collection[-1:], pkg_name)
 
+    if pkg_description:
```
pypingou: This actually changes the behavior we had, if pkg_description is ``' '`` it will return True while we want None in fact

pkgdb2/api/acls.py
```diff
@@ -143,7 +143,7 @@ def api_acl_update():
             SESSION.commit()
             output['output'] = 'ok'
             output['messages'] = messages
-        except pkgdblib.PkgdbException, err:
+        except pkgdblib.exceptions.PkgdbException, err:
```
pypingou: I wonder if it wouldn't be better to add a ``from pkgdb.lib.exceptions import PkgdbException`` at the top and use this

pkgdb2/lib/__init__.py
tyll: I do not follow. Before my change pkgdb would use the empty string if pkg_description is just a  space character:
```
In [3]: pkg_description = " "

In [4]: pkg_description.strip() if pkg_description else None
Out[4]: ''

```

pkgdb2/api/acls.py
tyll: If you prefer it, I will change it.

pypingou: Thanks

pkgdb2/lib/__init__.py
pypingou: That's a bug then, it should be: `` pkg_description.strip() if pkg_description and pkg_description.strip() else None``

tyll: Is this only needed for the description or ofr the other fields for the review URL and upstream URL as well?

pkgdb2/api/packages.py
```diff
@@ -152,7 +152,7 @@ def api_package_new():
             SESSION.commit()
             output['output'] = 'ok'
             output['messages'] = [message]
-        except pkgdblib.PkgdbException, err:
+        except pkgdblib.exceptions.PkgdbException, err:
```
pypingou: Should we use the same construct as in ``pkgdb2/api/acls.py``? It would allow us to be consistent :)

pkgdb2/lib/__init__.py
pypingou: All three fields apparently

tyll:
I believe I addressed all your suggestions now.

pkgdb2/api/admin.py
```diff
@@ -151,7 +151,7 @@ def api_admin_actions():
             status=status,
             count=True,
         )
-    except pkgdblib.PkgdbException, err:  # pragma: no cover
+    except pkgdblib.PkgdbException as err:  # pragma: no cover
```
pypingou: Here we use ``pkgdblib.PkgdbException`` but wasn't this the problem we were trying to address at first?

We adjusted ``api/acls.py`` to import the exception directly, should we adjust the others to behave in the same way?

pkgdb2/lib/__init__.py
```diff
@@ -1972,15 +1975,32 @@ def add_new_package_request(
     if pkg_collection.startswith(('el', 'epel')):
         _validate_pkg(session, pkg_collection[-1:], pkg_name)
 
+    # pkg_description, pkg_upstream_url and pkg_review_url should be None if
+    # empty
+    if pkg_description:
+        pkg_description = pkg_description.strip()
+    if not pkg_description:
+        pkg_description = None
```
pypingou: We could make this in two lines no?
````
if pkg_description:
        pkg_description = pkg_description.strip() or None
````

pkgdb2/api/admin.py
tyll: The problem was with ```PkgdbBugzillaException``` the ```PkgdbException``` is also availible in pkgdblib, because it is imported there. We can also make it the same in the other places, but it is going to extend the scope of this PR beyond its original goal.

pkgdb2/lib/__init__.py
tyll: yes, this seems to be possible but I am not sure that it is easier to read.

pkgdb2/api/admin.py
pypingou: I realize it is available in pkgdblib, but somehow it was failing in the tests despite that.
So, imho, either we are consistent and fix it everywhere or figure out more precisely why the tests were failing

pypingou:
I'm wondering if we shouldn't split this PR into two so that we can merge the fix to bugzilla URLs on one side and deal with the exceptions in another PR

pkgdb2/api/admin.py
tyll: > So, imho, either we are consistent and fix it everywhere or figure out more precisely why the tests were failing
￼There are two different exceptions, one with Bugzilla it is name. This one caused the test failure. The changes to the exception withouth Bugzilla in its name was there to make it more consistent.

tyll:
I hope I got all exceptions and made made it uniform. If there are no other changes IMHO a second PR makes it only more complicated without much gain.

pypingou:
> I hope I got all exceptions and made made it uniform. 

You got them all in the API but missed the imports in the UI module, do you want me to do it?

> If there are no other changes IMHO a second PR makes it only more complicated without much gain.

I'm fine with keeping it all in one PR, I was suggesting splitting as a way to not stalling these changes :)

pypingou:
>> I hope I got all exceptions and made made it uniform.

>You got them all in the API but missed the imports in the UI module, do you want me to do it?

Ok I must have had an old cache or so since I can now see the imports everywhere.

Thanks for this PR, sorry it turned out to be that big :)

pypingou:
:+1: to merge

tyll:
Thank you

PR #364, opened 2016-06-10
Let the Package return everything regardless of the namespace

pypingou:
Since we are able to filter by namespace, this will let Package return everything,
regardless of the namespaces.


sayanchowdhury:
LGTM :+1:

pypingou:
Thanks :)

PR #365, opened 2016-06-13
Do not hardcode the URLs to koji, bodhi, packages, bugz...

pypingou:


pkgdb2/templates/package.html
```diff
@@ -220,32 +220,41 @@ <h1 class="inline">
...
-          <a href="http://pkgs.fedoraproject.org/cgit/{{ package.name }}.git/"
+          <a href="{{ config['CGIT_URL'] % package.name }}"
           class="cvs">
           Package source</a>
       </li>
+      {% endif %}
+      {% if config.get('PKGS_BUG_URL') %}
       <li>
-          <a href="https://apps.fedoraproject.org/packages/{{ package.name }}/bugs"
+          <a href="{{ config['PKGS_BUG_URL'] % package.name }}/bugs"
```
sayanchowdhury: `/bugs` need to be removed here.

pypingou: Yup indeed, thanks :)

pkgdb2/templates/packager.html
```diff
@@ -11,9 +11,9 @@
   {{ packager | avatar | safe }}
   {{ packager }}</h1> (<a class="fas" target="_blank"
     {% if packager.startswith('group::') %}
-        href="https://admin.fedoraproject.org/accounts/group/view/{{ packager.replace('group::', '') }}"
+        href="{{ config['PKGDB2_FAS_URL'] }}/group/view/{{ packager.replace('group::', '') }}"
```
sayanchowdhury: Is it not required to update `PKGDB2_FAS_URL` in `default_config.py` ? It is set to `None` right now.

sayanchowdhury:
LGTM (y)

pkgdb2/templates/packager.html
pypingou: It's None right now but pkgdb will have more problem if it isn't set (like it won't be able to check if someone is a packager) so I am pretty confident that this will always be set.

pypingou:
Thanks for the review :)

PR #371, opened 2016-07-07
Expand the field fas_name from PackageListingAcl from 32 to 255 chars

pypingou:
FAS restrict usernames to 32 characters, so we took the same approach
and restricted our usernames to 32 characters but that was without
thinking about the groups which do not have that restriction on their
names. So far we were lucky but apparently it's time to take it into
account.


ignatenkobrain:
LGTM!

Issue #374, opened 2016-07-11
Allow blacklisting of names


puiterwijk:
We need to blacklist certain names in certain namespaces, like "fedora" in "docker".

This needs to be done before we allow requesting of "docker" namespaced items.

puiterwijk:
Note that we should perhaps also verify this in the admin API, just to make sure.

PR #376, opened 2016-07-18
Implement blacklisting feature

puiterwijk:
Fixes: #374

### DIFF
--- a/pkgdb2/default_config.py
+++ b/pkgdb2/default_config.py
@@ -52,6 +52,9 @@ SECRET_KEY = '<insert here your own key>'
 # List the ACL which are auto-approved (don't need reviewing)
 AUTO_APPROVE = ['watchcommits', 'watchbugzilla']
 
+### Blacklisted items
+REQUEST_BLACKLIST = []
+
 # List of FAS user that can be automatically approved w/o checking if they
 # are packagers
 AUTOAPPROVE_PKGERS = []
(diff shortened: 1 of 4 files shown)


pkgdb2/default_config.py
```diff
@@ -52,6 +52,9 @@
 # List the ACL which are auto-approved (don't need reviewing)
 AUTO_APPROVE = ['watchcommits', 'watchbugzilla']
 
+### Blacklisted items
+REQUEST_BLACKLIST = ['fedora']
```
pypingou: We probably want to make this a dict allowing to specify blacklisted package per namespace

pypingou:
Thanks for working on this, I've had it on my list so that makes my list shorted :)

utility/pkgdb2.cfg.sample
```diff
@@ -19,6 +19,9 @@ DB_URL='sqlite:////var/tmp/pkgdb2_dev.sqlite'
 ### pages
 ITEMS_PER_PAGE = 50
 
+### Blacklisted items
```
pypingou: Could we provide one or more example of what the content should look like? :)

puiterwijk: Done, one line below :smile:

pypingou: But maybe we could include it in the doc, something like:
~~~~
### Blacklisted items (following the format: ``namespage/packagename``)
~~~~
Thoughts?

pypingou:
Looks good to me, thanks!

Issue #375, opened 2016-07-16
new package default namespace should be better


nirik:
The "request new package" button brings up a form for requesting a new package. 
One of the fields is 'namespace'. It seems to now default to 'docker' where before it was 'rpm'.

Perhaps this could be sorted by number of packages in that namespace? 
If not, could we just always make rpm the top choice for now?

PR #377, opened 2016-07-22
Ensure the `rpms` namespace is always the first in the list (default)

pypingou:
Fixes https://github.com/fedora-infra/pkgdb2/issues/375

### DIFF
--- a/pkgdb2/ui/packages.py
+++ b/pkgdb2/ui/packages.py
@@ -1361,6 +1361,10 @@ def package_request_new():
         collections.append(collection)
 
     namespaces = pkgdblib.get_status(SESSION, 'namespaces')['namespaces']
+    # Ensure the `rpms` namespace is always the first in the list (the default)
+    if 'rpms' in namespaces:
+        namespaces.pop(namespaces.index('rpms'))
+        namespaces.insert(0, 'rpms')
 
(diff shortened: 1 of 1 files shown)


cverna:
:+1: for me

pkgdb2/ui/packages.py
```diff
@@ -1361,6 +1361,11 @@ def package_request_new():
         collections.append(collection)
 
     namespaces = pkgdblib.get_status(SESSION, 'namespaces')['namespaces']
+    default_ns = APP.config.get('DEFAULT_NAMESPACE', 'rpms')
+    # Ensure the `rpms` namespace is always the first in the list (the default)
+    if default_ns in namespaces:
+        namespaces.pop(namespaces.index(default_ns))
+        namespaces.insert(0, default_ns)
```
sayanchowdhury: Combining into one? `namespaces.insert(0, namespaces.pop(namespaces.index(default_ns)))`

sayanchowdhury:
GTG to me :+1:

pkgdb2/ui/packages.py
pypingou: It would work but I think I'll keep the two statements approach as it potentially will make it easier to debug if that ever needs to be

pypingou:
Thanks for the reviews :)

PR #381, opened 2016-07-26
Give some time indication when doing mass-branching

pypingou:


puiterwijk:
Perhaps also print the name of a package when it has been branched to see progress?

utility/pkgdb2_branch.py
```diff
@@ -185,6 +188,9 @@ def main():
     except pkgdb2.lib.PkgdbException, err:
         print err
         return 1
+    end = datetime.datetime.utcnow()
+    print "Finished at %s" % end
+    print "Last for %s" % (end - start)
```
puiterwijk: s/Last/Lasted/

pypingou:
> Perhaps also print the name of a package when it has been branched to see progress?

We can't do that anymore as the entire branching is now done via a two SQL queries which make the process much faster than iterating through the DB with python.

puiterwijk:
Ah, right. Fair enough. :+1:

puiterwijk:
:+1:

pypingou:
Thanks! :)

PR #384, opened 2016-07-30
Fix the location where the collection is converted into a product

pypingou:
Otherwise we run into KeyError exceptions for rather logical reasons


pypingou:
Adjusted in prod so that it works

pypingou:
Hm, not ready to merge, there is another fix incoming

puiterwijk:
Looks good to me :+1:

pypingou:
Thanks :)

PR #388, opened 2016-08-30
Change vagrant provisioning to use ansible

ryanlerch:
This changes the was we provision the vagrant devel
envrionment, from the shell provisioner to the ansible
one.

These changes are based heavily on the same work done by @bowlofeggs on bodhi in https://github.com/fedora-infra/bodhi/pull/893

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -7,29 +7,13 @@ Vagrant.configure(2) do |config|
   config.vm.network "forwarded_port", guest: 5000, host: 5001
   config.vm.synced_folder ".", "/vagrant", type: "sshfs"
 
-  config.vm.provision "shell", inline: "sudo dnf -y install python redhat-rpm-config python-devel postgresql-devel postgresql-server rpl python-alembic python-psycopg2 gcc"
+  # Ansible needs the guest to have these
+  config.vm.provision "shell", inline: "sudo dnf install -y libselinux-python python2-dnf"
 
-  config.vm.provision "shell", inline: "pip install kitchen paver urllib3"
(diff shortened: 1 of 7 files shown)


pypingou:
Looks good to me.

@bowlofeggs do you want to check it as well?

Vagrantfile
```diff
@@ -7,29 +7,13 @@ Vagrant.configure(2) do |config|
   config.vm.network "forwarded_port", guest: 5000, host: 5001
   config.vm.synced_folder ".", "/vagrant", type: "sshfs"
```
bowlofeggs: By the way, @dustymabe just got vagrant-sshfs into Bodhi today, so soon you won't need to set up the copr anymore. Pretty sweet, right?

devel/ansible/roles/dev/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
...
+      - redhat-rpm-config
+
+- name: Install python packages
+  pip:
+      name: "{{ item }}"
+      state: present
+  with_items:
+      - kitchen
+      - paver
+      - urllib3
```
bowlofeggs: I personally recommend just putting these packages with the ones above so it becomes one dnf transaction instead of two, only because it'll go a little faster. It's just a preference thing, so if you like this better feel free.

bowlofeggs:
I agree with @pypingou, this looks great. Feel free to merge as you please!

Vagrantfile
pypingou: \ó/

pypingou:
Let's merge, we can always revisit later to adjust :)

PR #389, opened 2016-08-31
Move the process of updating bugzilla on PoC changes to be fedmsg-based

pypingou:


pypingou:
While open this PR shouldn't be merged right now, I first would like to discuss with other users of pkgdb about this change and see if we shouldn't keep the current code (which hasn't be removed/deactivated in this changeset anyway) working with some sort of boolean in the configuration file to pick the method.

fedmsg_update_bz.py
```diff
@@ -0,0 +1,54 @@
...
+    def __init__(self, hub):
+        super(UpdateBZConsumer, self).__init__(hub)
+
+    def consume(self, msg):
+        msg = msg['body']
+        self.log.info("Got a message %r" % msg['topic'])
+
+        pkg_poc = msg['msg'].get('username')
+        if not pkg_poc:
+            msg['msg']['package_listing']['point_of_contact']
```
sayanchowdhury: This value is not assigned?

pypingou: Fixed :)

sayanchowdhury:
:+1:  for me.

pypingou:
I'm going to make a release without this change first, to be on the safe side :)

fedmsg_update_bz.py
```diff
@@ -0,0 +1,54 @@
...
+    def __init__(self, hub):
+        super(UpdateBZConsumer, self).__init__(hub)
+
+    def consume(self, msg):
+        msg = msg['body']
+        self.log.info("Got a message %r" % msg['topic'])
+
+        pkg_poc = msg['msg'].get('username')
+        if not pkg_poc:
+            pkg_poc = msg['msg']['package_listing']['point_of_contact']
```
bowlofeggs: When would pkg_poc have been None?

fedmsg_update_bz.py
```diff
@@ -0,0 +1,54 @@
...
+        pkg_prev_poc = msg['msg'].get('previous_owner')
+        pkg_name = msg['msg'].get('package_name')
+        if not pkg_name:
+            pkg_name = msg['msg']['package_listing']['package']['name']
```
bowlofeggs: When would pkg_name be None?

PR #391, opened 2016-09-01
Fix pkgdb2.api.extras to be more robust when extracting the summary of a bug

pypingou:


puiterwijk:
Looks sane to me.

pypingou:
Thanks :)

puiterwijk:
I have totally messed up this review :(.

puiterwijk:
Well, the change itself was sane, but the resulting code was not (But it also wouldn't work before this PR).

PR #395, opened 2016-09-13
update vagrant ansible role to upgrade database

ryanlerch:
Previously, the vagrant config did not upgrade the database
when provisioning the new machine. This could lead to issues
if a database change was commited to master, and we pull the
dbdump from prod.

This change copies the alembic.ini from the utility directory,
changes the dburl to what the vagrant setup expects, and runs
the alembic upgrade command.


pypingou:
Looks fine to me, we just need to be aware that this will only work on a db dump from prod, not on a db created locally from scratch.

ryanlerch:
yeah, the vagrant setup always pulls the prod db down when provisioning a new box, so we should be fine here (i think)

PR #396, opened 2016-09-19
Fix the runserver script so that APP is imported when setting the pofiler

pypingou:


sayanchowdhury:
Looks good to me :+1:

pypingou:
Thanks :)

PR #399, opened 2016-10-05
Update the README fixing the vagrant-sshfs parts

ryanlerch:
Previously the vagrant-sshfs plugin was only available for fedora via COPR. now it is in the main Fedora repos, the README is updated to insctruct new users to get it from there.


pypingou:
:+1: thanks!

pypingou:
Let's merge :)

PR #403, opened 2016-11-16
Add 'Koschei status' link.

voxik:
This link is displayed among other links to useful sites, such as pkgdb, cgit, bodhi, etc.

BTW1 I have not tested this, since I don't have the environment ...

BTW2 where the icons such as [1] comes from? Would be probably convenient to have icon_koschei.png there as well ....

BTW3 when I shown this PR to @mizdebsk, he pointed out that similar functionality is available in FedMenu, but honestly, although I know that FedMenu exists, I had no idea that I can found such link there. I suppose I am not the only exception. But if this gets rejected due to this, then I'm going to send another PR removing all the pkgdb, cgit, bodhi, etc icons from the top ;)

[1]: https://fedoraproject.org/static/images/icons/fedora-infra-icon_bodhi.png


kjn:
:+1: Would definitely be nice to have.

mizdebsk:
:+1: Would definitely be nice to have.

pypingou:
Looks all good to me, thanks! :)

PR #404, opened 2016-12-26
Allow not providing a review_url when un-retiring master

pypingou:


puiterwijk:
This decision has been reverted? As this looks pretty specific code to me.

pypingou:
Yes, as of fedora-infra/packagedb-cli#98 pkgdb-cli is able to check for how long a package was retired.

Before we enforced the re-review URL all the time, because we were not able to enforce it only when the package had been retired for less than 2 weeks.
With this commit we're dropping the requirement on pkgdb itself and it will be the pkgdb admins who will  check and report if the request is invalid.

puiterwijk:
Ah, okay. I think I'd prefer it if this check would happen on the server, but if only admins run it I'm +1.

pypingou:
Ok, I'm starting to think let's just fly with this for now and reconsider if it causes problems.

Note: I'm going to wait a bit before making that pkgdb2 release to pkgdb-admin to land to at least -testing, preferably updates

PR #405, opened 2017-01-06
Fix approving a new package request in the docker namespace

pypingou:
When we approve a new package request, in any namespace, we normally go
then through all the other pending requests to automatically create the
branch (for the branches for which we can).
With non-rpm namespaces, this logic was broken since we were not including
the namespace in the search for other requests and we were ending up
with an exception saying that there are no such package.
This was confusing when processing the requests via pkgdb-admin since it
would return an error (not even a clear one) and then when checking the
UI, the new package was created.

This commit fixes that situation.


pypingou:
This PR has been already applied/hotfixed in prod.

maxamillion:
LGTM 👍 

Thanks for the quick fix!

pypingou:
Thanks for the review :)

Issue #378, opened 2016-07-22
fedmsg for pkgdb when a package is approved for unretirement


mohanboddu:
We are writting a script which will unblock a package in Koji if it is approved for unretirement in pkgdb. It would be really helpful if a message is published when it does happens.

pypingou:
I would actually expect https://apps.fedoraproject.org/datagrepper/raw?topic=org.fedoraproject.prod.pkgdb.package.update.status to cover also unretire, odd that it doesn't :(

keszybz:
That sounds like a good idea. Maybe it's a matter of just adding one right line somewhere in pkgdb?

pypingou:
Fixed in the proposed PR :)

PR #407, opened 2017-01-17
Fix sending notifications when we update the status of a package on a branch

pypingou:
Turns out we were doing pretty bad at announcing when we changed the
status of a package on a certain branch (ie: going from Approved to
Orphaned or Retired and the other way around).
This commit fixes this.

Fixes https://github.com/fedora-infra/pkgdb2/issues/378


keszybz:
Beautiful, thanks!

pypingou:
Thanks for the review :)

PR #409, opened 2017-01-18
Fix bug: missing variable `status` -- thanks unit-tests

pypingou:
Turns out we need a `status` variable to send the fedmsg notif about
status change, so let's have one so that our code compiles and actually
works.


pypingou:
Thanks for the reviews :)

PR #413, opened 2017-04-20
Update documentation to refer to pagure.io instead of trac

voxik:
BTW, if there was a way to open pagure ticket with prefilled subject and body, that would be sweet ...

### DIFF
--- a/doc/groups.rst
+++ b/doc/groups.rst
@@ -20,10 +20,9 @@ One requirement for the mailing list address:
   bugzilla account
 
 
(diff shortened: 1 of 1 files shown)


pypingou:
> BTW, if there was a way to open pagure ticket with prefilled subject and body, that would be sweet ...

Such as : https://docs.pagure.org/pagure/usage/tips_tricks.html#pre-fill-issue-template-using-the-url ? :)

pypingou:
Change looks fine, thanks! :)

PR #414, opened 2017-04-27
Document suggested setup of package maintained by group.

voxik:
This is attempt to provide more insight how to setup the package ACL when it should be maintained by FAS group.

### DIFF
--- a/doc/groups.rst
+++ b/doc/groups.rst
@@ -54,8 +54,4 @@ The suggested package configuration for groups maintenance is:
 * Group should have ``watchbugzilla`` permissions to allow group to be
   notified about bugzilla issues associated with the package.
 * Optionally, the group can become PoC of the package. This effectively
-  results only in change of default assignee in bugzilla, nothing else. Keep
-  in mind, that this does not replace ``watchbugzilla`` bugzilla presmissions,
-  since the default assignee is not in CC list of the ticket by default
-  and therefore once the assignee in bugzilla was changed, the the group would
-  not be notified about the changes in bugzilla ticket anymore.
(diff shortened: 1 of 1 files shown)


doc/groups.rst
```diff
@@ -36,3 +40,22 @@ which ACL you want to give and on which branch.
...
+* Group should have ``commits`` bit. This allows every member of the
+  group to make changes to package in dist-git. This is actually the main
+  purpose of groups.
+* Add "watchcommits" bit to allow notification of the group members about
+  changes in the package dist-git.
+* Group should have ``watchbugzilla`` permissions to allow group to be
+  notified about bugzilla issues associated with the package.
+* Optionally, the group can become PoC of the package. This effectively
+  results only in change of default assignee in bugzilla, nothing else. Keep
+  in mind, that this does not replace ``watchbugzilla`` bugzilla presmissions,
```
pypingou: Typo on the last word here: presmissions :)

doc/groups.rst
```diff
@@ -36,3 +40,22 @@ which ACL you want to give and on which branch.
...
+  since the default assignee is not in CC list of the ticket by default
```
pypingou: Do we double-check this? I believe it is true but I might have been wrong

voxik:
1. It seems I was not right about the "default CC", so I dropped the note.
2. I fixed some quotes.
3. I added another bullet suggesting to cleanup the ACLs, since users don't need to be listed anywhere, if they are already member of some group with appropriate rights.

Let me know if I should squash the commits.

pypingou:
Looks all good to me, thanks! :)

PR #415, opened 2017-05-22
Add a deprecation banner

mprahl:
This PR adds a red banner warning users that PkgDB will be replaced. This PR may not make sense to merge, as just a patch in production might do, but it should be here for discussion.

Here is a screenshot of the change.
Banner

### DIFF
--- a/doc/groups.rst
+++ b/doc/groups.rst
@@ -54,8 +54,4 @@ The suggested package configuration for groups maintenance is:
 * Group should have ``watchbugzilla`` permissions to allow group to be
   notified about bugzilla issues associated with the package.
 * Optionally, the group can become PoC of the package. This effectively
-  results only in change of default assignee in bugzilla, nothing else. Keep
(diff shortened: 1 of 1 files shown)


pkgdb2/templates/master.html
```diff
@@ -89,6 +89,16 @@ <h4 class="hide">Main Site Links:</h4>
           {% endif %}
         </span>
 
+        <div id="deprecationWarningDiv">
+            <p>
+                Attention! PkgDB will be replaced by June xx, 2017.
```
mprahl: This will need an official date as we finalize the timeline.

pypingou:
Yeah, not sure we want to merge this :)

ralphbean:
Sent FBR for this to the infra list:  https://lists.fedoraproject.org/archives/list/infrastructure@lists.fedoraproject.org/thread/5COJNMHCWD3QM7PP6Q4J2NBNZ5UN4E4A/#5COJNMHCWD3QM7PP6Q4J2NBNZ5UN4E4A

PR #416, opened 2017-07-26
Add support for namespace to the /api/notify endpoint

pypingou:
Signed-off-by: Pierre-Yves Chibon <pingou@pingoured.fr>


pkgdb2/api/extras.py
```diff
@@ -262,11 +275,14 @@ def api_notify():
     :kwarg eol: Set to True if you want to include end of life
         distributions
     :kwarg format: Specify if the output if text or json.
+    :kwarg namespace: Specify if the namespace should be included in the
+        output (defaults to False).
```
puiterwijk: Why do you make this an option?
It's just another key in a dict, so it should be fine to just always add it, which also makes it easier to test for correctness.

pypingou: Because that would change the output and thus break tools relying on this format

puiterwijk: ok

pypingou:
Thanks! :)

PR #417, opened 2017-07-26
Fix running the query on postgresql

pypingou:
The identifier must be part of the group_by on postgresql otherwise
it fails.

Signed-off-by: Pierre-Yves Chibon <pingou@pingoured.fr>


pypingou:
This has been hotfixed in prod

